PR #2, opened 2018-10-31
Fix DSLogon LOA logic

patrickvinograd:
Eliminates broken parsing logic, still testing DS Logon but this should at least fix the ID.me path

### DIFF
--- a/src/routes/handlers.js
+++ b/src/routes/handlers.js
@@ -1,3 +1,4 @@
+import { IDP_SSO } from "./constants";
 import { getPath, getReqUrl } from "../utils";
 import assignIn from "lodash.assignin";
 import SessionParticipants from "samlp/lib/sessionParticipants";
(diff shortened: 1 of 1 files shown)


src/routes/handlers.js
```diff
@@ -194,10 +195,10 @@ export const acsFactory = (app, acsUrl) => {
     function(req, res, next) {
       console.log(req.user);
       if (req.user && req.user.claims &&
-          (req.user.claims.level_of_assurance != '3' ||
-           req.user.claims.dslogon_assurance.parseInt() < 2)) {
+          (req.user.claims.level_of_assurance != '3' &&
+           parseInt(req.user.claims.dslogon_assurance) != 2)) {
```
edpaget: Could this just be a `!= '2'` instead of needing to parse int. 

Also this and one the loa test above should probably be `!==` to be on the javascript safe side. My mistake on the first one.

patrickvinograd: Fixed

patrickvinograd:
DSLogon now fails on the way back to Okta due to inability to JIT create a profile, so some aspect of profile mapping is not working. But we can go ahead and merge/deploy this while we debug that.

PR #3, opened 2018-10-31
Add DSLogon claims in config

patrickvinograd:
This allows the proxy to map through dslogon-specific claims through to Okta, and then there is corresponding config on the Okta side to map them through to the user profile.


patrickvinograd:
I pushed this to one of the dev saml proxy instances and was able to complete DSLogon sign-in

PR #10, opened 2018-11-13
MHV support

patrickvinograd:
- Adds profile mapping to support MHV.
- Adds a profile mapping option that supports a transformation method, used to map MHV's complex profile object onto a single account_type string.
- Updates allowed LOAs for DSLogon
- Preserves ID.me-provided authnContext in response path through to Okta.


patrickvinograd:
Tested interactively in dev.

PR #15, opened 2018-12-07
Add oauth proxy to repo

edpaget:
Add oauth proxy and reorganizes repo to handle. Also adds healthchecks to docker containers, so we can create a third container on deploy to check status of the proxies and report to ELB healthcheck.


oauth-proxy/index.js
```diff
@@ -0,0 +1,124 @@
...
+      state = document.state.S;
+      await dynamoClient.saveToDynamo(dynamo, state, 'refresh_token', tokens.refresh_token);
+    } else if (req.body.grant_type === 'authorization_code') {
+      tokens = await client.grant(
+        {...req.body, redirect_uri }
+      );
+      const document = await dynamoClient.getFromDynamoBySecondary(dynamo, 'code', req.body.code);
+      state = document.state.S;
+      await dynamoClient.saveToDynamo(dynamo, state, 'refresh_token', tokens.refresh_token);
+    } else {
```
dvogel: In the redirect route above, if `saveToDynamo` fails then the lookup here will also fail. IIUC once we receive the redirect request it's a little late to stop the train so we have to send the user on. Maybe the best we can do it add logging to both save path and this path?

oauth-proxy/index.js
```diff
@@ -0,0 +1,124 @@
...
+  } catch (error) {
+    console.error(error);
+    process.exit(1);
+  }
+})();
+
+module.exports = {
+  createIssuer,
+  startApp,
+}
```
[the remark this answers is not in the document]
edpaget: No these were just from me testing things out in the node repl.

oauth-proxy/index.js
edpaget: :+1: I'll make sure to rescue from all the dynamo calls. They're only required by one consumer (returning state like this doesn't seem to be in any of the oauth, openid, or smart of fhir specs), so a failure to look up should just be caught and logged like you suggest.

edpaget:
Still need to add in the check that the initial redirect URL we're sent is allowed, will work on that.

patrickvinograd:
I think the Liberty team is hosting a SMART capabilities document under their `/metadata` endpoint in lieu of the smart-configuration.json, but I'm not opposed to keeping place what I added there, though I also want to proxy the openid-configuration.json.

edpaget:
Ah I think I missed that that the smart config was still there in a different commit it looks all set to me.

edpaget:
Added a check that hits okta's api to lookup whitelisted redirect URIs for applications, I also rearranged the config format a bit.

PR #18, opened 2018-12-11
Ux for non loa3 users

charleystran:
This is a pull request to direct users to a new page explaining they need identity verification if they are not loa3. https://github.com/department-of-veterans-affairs/vets-contrib/issues/609


edpaget:
Not sure how users get directed to page?

charleystran:
hey @edpaget you should be able to get there by using DSlogon with the loa1 user here https://github.com/department-of-veterans-affairs/vets.gov-team/blob/master/Products/Identity/reference_documents/ds%20logon/ds-logon-lower-env-test-accounts.md

saml-proxy/src/routes/constants.ts
```diff
@@ -5,3 +5,4 @@ export const IDP_SIGN_IN = '/samlproxy/idp/signin';
 export const SP_LOGIN_URL ='/samlproxy/sp/login';
 export const SP_METADATA_URL = '/samlproxy/sp/metadata';
 export const SP_ERROR_URL = '/samlproxy/sp/error';
+export const SP_NOT_SUFFICIANT_LOA = '/samlproxy/sp/not_sufficiant_loa';
```
patrickvinograd: I would name this route "/samlproxy/sp/idproof" or "/samlproxy/sp/verify". And optimally rename the constant and references to it as well so there's not a mismatch.

saml-proxy/src/routes/handlers.js
```diff
@@ -174,7 +174,7 @@ const processAcs = (acsUrl) => [
     if (req.user && req.user.claims &&
         !sufficientLevelOfAssurance(req.user.claims)) {
       res.redirect(url.format({
-        pathname: IDP_SSO,
+        pathname: SP_NOT_SUFFICIANT_LOA,
         query: {
```
patrickvinograd: Part of what I'm not sure of (and what will be hard to test) is whether if somebody does accept our prompt to identity proof, if they will then wind up back at the originating application. In addition to the authnContext, for this to work we'd need to pass the `RelayState` query param through the ID proofing flow. It gets generated by Okta, and propagated through the entire SAML flow through our proxy to ID.me and back to Okta.

patrickvinograd: I think the authnContext here is redundant since the idme_login_link and signup_link imply it, but we do need to make sure RelayState is sent, whatever the best way of doing that is.

charleystran: I can remove the authContext here. I dont think it is used, but I may need some guidance(especially for testing) on how to re-pass the relayState. I **thought** that was in the session, but if it is not, we will need to pass through in the url.

saml-proxy/src/routes/constants.ts
charleystran: I went with verify. I pushed and updated the name of the view as well.

saml-proxy/src/routes/handlers.js
charleystran: hey @patrickvinograd It was not properly redirecting, but I was able to update it to do so.

PR #24, opened 2018-12-14
Set HTTP keepalive timeouts greater than ELB idle timeouts

patrickvinograd:
Per guidance here: https://docs.aws.amazon.com/elasticloadbalancing/latest/classic/config-idle-timeout.html

Default was 5 seconds, going to set ELB idle timeout to 60 seconds.


patrickvinograd:
Sorry, going to set the ELB timeout to 60s, and the node app HTTP keepalive to 75s, in line with the recommendations in the AWS link above.

PR #34, opened 2019-04-01
Ensure docker stage is run on jenkins swarm node

kalilsn:
This second `agent` directive was causing our builds to sometimes change over to the Jenkins master node, which doesn't have docker installed.

Context:
https://lighthouseva.slack.com/archives/CE79R5SR1/p1554141205032100
https://dsva.slack.com/archives/C7S6EA0ES/p1554142651056000

### DIFF
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -19,6 +19,7 @@ pipeline {
         dockerfile {
           args "--entrypoint='' -u 0:0"
           dir "saml-proxy"
+          label 'vetsgov-general-purpose'
         }
       }
 


jasalisbury:
Didn't catch it at first but looks like the other strings in that file are double quotes, not single quotes

kalilsn:
I'm seeing a mixture throughout 🙃. It doesn't matter here, and if I start fixing things like that then I'm going to spend all day refactoring our Jenkinsfiles, so I'd like to leave it as is.

PR #36, opened 2019-04-09
Look up MVI users with OpenidAuth::V0::MviUsersController POST action…

mjlumetta:
… for UTF-8 - 1380

## Overview 

Completes [1380](https://github.com/department-of-veterans-affairs/vets-contrib/issues/1380) on the SAML proxy side.

Changes `VetsAPIClient.ts` to use the `mvi_users#search` POST action instead of the `mvi_users#show` GET action to look up users in MVI.

This depends on the [corresponding Vets API PR](https://github.com/department-of-veterans-affairs/vets-api/pull/2945).

## Testing

To test, make sure that you have the correct vets-api version. If the PR linked above hasn't merged, you should checkout `origin/ml-1380-utf8-header-error`.

* Verify that you can sign into a sample app locally using OAuth with this branch of the SAML proxy and the above branch of the Vets API with user va.api.user+idme.103@gmail.com.

You can't really determine for certain that the error doesn't occur unless you're debugging with byebug or console output as a developer because it no longer crashes anyway, but it should work as expected. I've tested that the output looks as expected in the server, and other devs are welcome to do the same.

### DIFF
--- a/saml-proxy/src/VetsAPIClient.ts
+++ b/saml-proxy/src/VetsAPIClient.ts
@@ -27,24 +27,29 @@ export class VetsAPIClient {
   public async getMVITraitsForLoa3User(user: SAMLUser) : Promise<{ icn: string, first_name: string, last_name: string }> {
     const headers = {
       'apiKey': this.token,
-      'x-va-idp-uuid': user.uuid,
-      'x-va-user-email': user.email,
-      'x-va-dslogon-edipi': user.edipi || null,
-      'x-va-mhv-icn': user.icn || null,
-      'x-va-ssn': user.ssn || null,
-      'x-va-first-name': user.firstName || null,
-      'x-va-middle-name': user.middleName || null,
-      'x-va-last-name': user.lastName || null,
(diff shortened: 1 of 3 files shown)


kalilsn:
Even if the vets-api PR does get merged today, let's not merge this one since the saml-proxy deploy happens a few hours earlier! This looks good though, I'll approve it tomorrow.

saml-proxy/src/VetsAPIClient.ts
```diff
@@ -27,24 +27,29 @@ export class VetsAPIClient {
...
+      'first_name': user.firstName || null,
+      'middle_name': user.middleName || null,
+      'last_name': user.lastName || null,
+      'dob': user.dateOfBirth || null,
+      'gender': user.gender || null,
+      'level_of_assurance': '3',
+    };
+
     // @ts-ignore TS7017
     Object.keys(headers).forEach((key) => (headers[key] == null) && delete headers[key]);
```
dvogel: Now that the `headers` object has been simplified to contain only the API key, we can remove this line that deletes null headers.

PR #38, opened 2019-04-24
Attempt VSO Search If MVI Fails

ATeal:
In order to support VSO users in the oauth flow, we need to run a check to see if they exist if MVI fails. This should allow VSO uses to flow through like normal.

resolves https://github.com/department-of-veterans-affairs/vets-contrib/issues/1823

This is dependent on: https://github.com/department-of-veterans-affairs/vets-api/pull/2991 && https://github.com/department-of-veterans-affairs/vets-contrib/issues/1863

### DIFF
--- a/saml-proxy/src/VetsAPIClient.ts
+++ b/saml-proxy/src/VetsAPIClient.ts
@@ -56,7 +56,7 @@ export class VetsAPIClient {
 
   public async getVSOSearch(firstName: string, lastName: string) : Promise<{poa: string}> {
 
-    const body = {
+    const qs = {
       'first_name': firstName,
       'last_name': lastName
     }
@@ -65,8 +65,8 @@ export class VetsAPIClient {
       url: `${this.apiHost}${VSO_SEARCH_PATH}`,
       json: true,
       headers: this.headers,
-      body,
+      qs,
     });
-    return response.data.attributes;
+    return response.data.attributes.power_of_attorney_code;
   }
 }
--- a/saml-proxy/src/VetsAPIClient.ts
+++ b/saml-proxy/src/VetsAPIClient.ts
@@ -54,7 +54,7 @@ export class VetsAPIClient {
     return response.data.attributes;
   }
 
-  public async getVSOSearch(firstName: String, lastName: String) : Promise<{poa: string}> {
+  public async getVSOSearch(firstName: string, lastName: string) : Promise<{poa: string}> {
 
     const body = {
       'first_name': firstName,
--- a/saml-proxy/src/VetsAPIClient.ts
+++ b/saml-proxy/src/VetsAPIClient.ts
@@ -14,7 +14,7 @@ export interface SAMLUser {
 }
 
 const MVI_PATH = '/internal/auth/v0/mvi-user';
-const VSO_SEARCH_PATH = '/services/veteran/v0/representatives/search';
+const VSO_SEARCH_PATH = '/services/veteran/v0/representatives/find_rep';
 
 export class VetsAPIClient {
   token: string;
@@ -67,6 +67,6 @@ export class VetsAPIClient {
       headers: this.headers,
       qs,
     });
-    return response.data.attributes.power_of_attorney_code;
+    return response.data.attributes
   }
 }
--- a/saml-proxy/src/VetsAPIClient.ts
+++ b/saml-proxy/src/VetsAPIClient.ts
@@ -13,7 +13,7 @@ export interface SAMLUser {
   icn?: string;
 }
 
-const LOOKUP_PATH = '/internal/auth/v0/mvi-user';
+const MVI_PATH = '/internal/auth/v0/mvi-user';
 
 export class VetsAPIClient {
   token: string;
@@ -44,7 +44,7 @@ export class VetsAPIClient {
     };
 
     const response = await request.post({
-      url: `${this.apiHost}${LOOKUP_PATH}`,
+      url: `${this.apiHost}${MVI_PATH}`,
       json: true,
       headers,
       body,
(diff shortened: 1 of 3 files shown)


saml-proxy/src/routes/acsHandlers.ts
```diff
@@ -61,7 +61,15 @@ export const loadICN = async (req: IConfiguredRequest, res: Response, next: Next
...
   } catch (error) {
-    res.render(mviErrorTemplate(error), {});
+    const mviError = error;
+
+    try  {
+      const { poa } = await req.vetsAPIClient.getVSOSearch(req.user.claims.firstName, req.user.claims.lastName);
+      req.user.claims.poa = poa;
+      next();
+    } catch (error) {
+      res.render(mviErrorTemplate(mviError), {});
```
kalilsn: Is it appropriate to show the MVI error to a VSO user? I think there's a case here that's going unhandled: VSO user attempts to authenticate, but the getVSOSearch call fails for some reason (upstream service is down, network error, etc.).

saml-proxy/src/VetsAPIClient.ts
```diff
@@ -44,9 +46,25 @@ export class VetsAPIClient {
...
+    });
+    return response.data.attributes;
+  }
+
+  public async getVSOSearch(firstName: string, lastName: string) : Promise<{poa: string}> {
+
+    const body = {
+      'first_name': firstName,
+      'last_name': lastName
+    }
```
kalilsn: I think these should be query parameters, not a body, since this is a GET request (and because that's what the endpoint [appears to support](https://github.com/department-of-veterans-affairs/vets-api/pull/2991/files#diff-ce96abb4a2c87e01bf2551f7a1d056b7R10))

saml-proxy/src/VetsAPIClient.ts
```diff
@@ -44,9 +46,25 @@ export class VetsAPIClient {
...
+
+    const response = await request.get({
+      url: `${this.apiHost}${VSO_SEARCH_PATH}`,
       json: true,
-      headers,
+      headers: this.headers,
       body,
     });
     return response.data.attributes;
```
kalilsn: I think this needs to be `response.data.power_of_attorney_code`

saml-proxy/src/routes/acsHandlers.ts
ATeal: So I actually did some checks on this, it renders something like this
image (4)

Which doesn't specifically mention mvi. It's unclear how we should handle the error because really VSO is just a fall back and probably the more nice case right now. Open to suggestions for this though

saml-proxy/src/VetsAPIClient.ts
ATeal: Yeah rails is ultra nice and will pick up body or query params into the params block. You're probably right though these should be query params. I'll change this!

saml-proxy/src/VetsAPIClient.ts
ATeal: Mmmm thanks for calling attention to this. Potentially may be changing on the rails side, I'll update this accordingly

saml-proxy/src/routes/acsHandlers.ts
dvogel: I think it's worth renaming that function to something like `unknownUsersErrorTemplate`

saml-proxy/src/VetsAPIClient.ts
```diff
@@ -13,21 +13,23 @@ export interface SAMLUser {
   icn?: string;
 }
 
-const LOOKUP_PATH = '/internal/auth/v0/mvi-user';
+const MVI_PATH = '/internal/auth/v0/mvi-user';
+const VSO_SEARCH_PATH = '/services/veteran/v0/representatives/search';
```
dvogel: Is the representative search going to be a publicly available API? If not I think it should probably be hosted at `/internal/...` like the MVI lookup endpoint.

saml-proxy/src/VetsAPIClient.ts
```diff
@@ -44,11 +46,27 @@ export class VetsAPIClient {
...
+    }
+
+    const response = await request.get({
+      url: `${this.apiHost}${VSO_SEARCH_PATH}`,
+      json: true,
+      headers: this.headers,
+      qs,
+    });
+    return response.data.attributes.power_of_attorney_code;
+  }
```
dvogel: Does this pass the type checker? It looks like it's expected to return a promise of a hash but instead if returns a promise of a string.

saml-proxy/src/routes/acsHandlers.ts
```diff
@@ -61,7 +61,15 @@ export const loadICN = async (req: IConfiguredRequest, res: Response, next: Next
     req.user.claims.lastName = last_name;
     next();
   } catch (error) {
-    res.render(mviErrorTemplate(error), {});
+    const mviError = error;
+
+    try  {
+      const { poa } = await req.vetsAPIClient.getVSOSearch(req.user.claims.firstName, req.user.claims.lastName);
+      req.user.claims.poa = poa;
+      next();
```
dvogel: It seems like we can skip setting `req.user.claims.poa` since it's not used elsewhere. The main purpose of adding user claims it to ensure they end up in the user's okta profile. We probably don't want to do that since there's no natural hook for it to be updated.

saml-proxy/src/VetsAPIClient.ts
ATeal: It's internal only, but I'm happy to communicate that to the team.

saml-proxy/src/VetsAPIClient.ts
ATeal: grrr, originally I had it as `response.data.attributes` but I think I had it messed up because of the field was remapped from `poa -> power_of_attorney_code` but I think that change got changed (lol words amirite) either way I'll ensure this is fixed before merging.

saml-proxy/src/routes/acsHandlers.ts
ATeal: Yeah we can totally remove that. Mostly added that to follow suit. I'll just leave it at the await.

PR #39, opened 2019-05-23
Add idp param from config if not set in authorize request

kalilsn:
https://github.com/department-of-veterans-affairs/vets-contrib/issues/2066

Adds idp param to authorize redirect if it's not already present.

### DIFF
--- a/oauth-proxy/cli.js
+++ b/oauth-proxy/cli.js
@@ -54,7 +54,7 @@ function processArgs() {
         description: 'apiKey permitting access to validate endpoint',
       },
       idp: {
-        description: 'Okta IDP identifier to be added as ?idp param if not specified by user in authorize request'
+        description: 'Okta IDP identifier to be added as a query param (idp) if not specified by user in authorize request'
       },
     })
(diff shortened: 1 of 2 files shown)


oauth-proxy/cli.js
```diff
@@ -52,7 +52,10 @@ function processArgs() {
       },
       validate_apiKey: {
         description: 'apiKey permitting access to validate endpoint',
-      }
+      },
+      idp: {
+        description: 'Okta IDP identifier to be added as ?idp param if not specified by user in authorize request'
```
jasalisbury: > ?idp param if not specified by

Did you mean to have `?idp` instead of `idp`?

kalilsn: Yes it was intentional, but I'm very open to any suggestion for changing it! I was just trying to indicate that it's a query parameter in an http request. Any ideas for how I could make this message clearer?

jasalisbury: Don't think it's any better than what you already have, `Okta IDP identifier to be added as a query param if not specified by user in authorize request`

PR #40, opened 2019-06-08
Use req.flash properly in passport.authenticate()

mjlumetta:
Solves an error that existed in the SAML Proxy but wasn't exposed because it required a bad signature in the staging env (in the investigation of [vets-contrib #2069](https://github.com/department-of-veterans-affairs/vets-contrib/issues/2069)) to expose the bug. So, generally speaking, I don't think that we need to worry about this happening to an end user, unless there is another way to cause it. The reason is that ID.me will handle actual authentication errors like bad email/password combination - it never sends a `SAMLResponse` back to the SAML Proxy with a failure message to my knowledge in a successfully configured environment. Here's what ID.me does if you enter a bad password or username: 

IDmeLoginError

Alternative solution would be to remove `this.failureFlash = true` in `SPConfig`.

There's also a lot wrong with this currently, as this fix doesn't actually make the error handler comprehensible. It still 404s because `passport.authenticate` then immediately tries to redirect to `SP_ERROR_URL`, which we specified via `this.failureRedirect` in `SPConfig`. `SP_ERROR_URL` doesn't have a route handler, though, so we should probably decide what we want that backup to look like in case something is wrong in a deployed env.

### DIFF
--- a/saml-proxy/package-lock.json
+++ b/saml-proxy/package-lock.json
@@ -1942,6 +1942,11 @@
         "xdg-basedir": "^3.0.0"
       }
     },
+    "connect-flash": {
+      "version": "0.1.1",
+      "resolved": "https://registry.npmjs.org/connect-flash/-/connect-flash-0.1.1.tgz",
+      "integrity": "sha1-2GMPJtlaf4UfmVax6MxnMvO2qjA="
+    },
     "console-browserify": {
       "version": "1.1.0",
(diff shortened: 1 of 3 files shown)


mjlumetta:
>> so we should probably decide what we want that backup to look like in case something is wrong in a deployed env.

> Are you able to demo that? It's probably worth showing in stand-up so everyone can get on the same page.

@dvogel yes can do

PR #42, opened 2019-06-12
Bootstrap test framework and add an end-to-end test for basic coverage.

dvogel:
I started adding CORS support to the oauth-proxy and realized there's no test suite. This adds a few tests to get minimal coverage.

```
-----------------------------|----------|----------|----------|----------|-------------------|
File                         |  % Stmts | % Branch |  % Funcs |  % Lines | Uncovered Line #s |
-----------------------------|----------|----------|----------|----------|-------------------|
All files                    |    58.82 |    32.61 |    59.09 |    59.11 |                   |
 oauth-proxy                 |    52.63 |    30.95 |    41.38 |    52.94 |                   |
  cli.js                     |    66.67 |      100 |        0 |    66.67 |                 4 |
  dynamo_client.js           |        8 |        0 |        0 |        8 |... 88,89,90,91,93 |
  index.js                   |    60.14 |    38.24 |    66.67 |    60.56 |... 21,322,324,325 |
 oauth-proxy/tests           |    90.91 |       50 |    93.33 |    90.91 |                   |
  backgroundServer.js        |    77.78 |       50 |      100 |    77.78 |             13,20 |
  upstreamOAuthTestServer.js |    95.83 |      100 |    90.91 |    95.83 |               167 |
-----------------------------|----------|----------|----------|----------|-------------------|
```

The coverage isn't great right now but this feels like a good compromise between coverage and improbability. There's a bit of a chicken and egg problem here. The app is currently written as one giant function. As such, it resists most testing approaches. Yet, since there's no tests it's hard to change the app to be more amenable to testing. My goal with these tests is to establish some coverage from a high level, fully-integrated testing approach so that we can start breaking the app up into a few separate modules. Those modules will be more amenable to more isolated testing approaches, which will allow us to boost coverage in the future.

### DIFF
--- a/oauth-proxy/index.js
+++ b/oauth-proxy/index.js
@@ -11,13 +11,6 @@ const okta = require('@okta/okta-sdk-nodejs');
 const morgan = require('morgan');
 const requestPromise = require('request-promise-native');
 
-const config = processArgs();
-const oktaClient = new okta.Client({
-  orgUrl: config.okta_url,
-  token: config.okta_token,
-  requestExecutor: new okta.DefaultRequestExecutor()
-});
-const { well_known_base_path } = config;
 const appRoutes = {
   authorize: '/authorization',
   token: '/token',
@@ -25,14 +18,6 @@ const appRoutes = {
   introspection: '/introspect',
   jwks: '/keys',
   redirect: '/redirect'
-};
-const redirect_uri = `${config.host}${well_known_base_path}${appRoutes.redirect}`;
-const metadataRewrite = {
-  authorization_endpoint: `${config.host}${well_known_base_path}${appRoutes.authorize}`,
-  token_endpoint: `${config.host}${well_known_base_path}${appRoutes.token}`,
-  userinfo_endpoint: `${config.host}${well_known_base_path}${appRoutes.userinfo}`,
-  introspection_endpoint: `${config.host}${well_known_base_path}${appRoutes.introspection}`,
-  jwks_uri: `${config.host}${well_known_base_path}${appRoutes.jwks}`,
 };
 const openidMetadataWhitelist = [
   "issuer",
@@ -71,23 +56,26 @@ const smartCapabilities = [
   "permission-patient",
 ]
 
-const dynamo = dynamoClient.createClient(
-  Object.assign({},
-    { region: config.aws_region },
-    config.aws_id === null ? null : { accessKeyId: config.aws_id },
-    config.aws_secret === null ? null : { secretAccessKey: config.aws_secret }
-  ),
-  config.dynamo_local,
-  config.dynamo_table_name,
-);
-
-async function createIssuer() {
+async function createIssuer(config) {
   return await Issuer.discover(config.upstream_issuer);
 }
 
-function startApp(issuer) {
+function buildMetadataRewriteTable(config, appRoutes) {
+  return {
+    authorization_endpoint: `${config.host}${config.well_known_base_path}${appRoutes.authorize}`,
+    token_endpoint: `${config.host}${config.well_known_base_path}${appRoutes.token}`,
+    userinfo_endpoint: `${config.host}${config.well_known_base_path}${appRoutes.userinfo}`,
+    introspection_endpoint: `${config.host}${config.well_known_base_path}${appRoutes.introspection}`,
+    jwks_uri: `${config.host}${config.well_known_base_path}${appRoutes.jwks}`,
+  };
+}
+
+function buildApp(config, issuer, oktaClient, dynamo, dynamoClient) {
+  const { well_known_base_path } = config;
+  const redirect_uri = `${config.host}${well_known_base_path}${appRoutes.redirect}`;
+  const metadataRewrite = buildMetadataRewriteTable(config, appRoutes);
+
   const app = express();
-  const { port } = config;
   const router = new express.Router();
   app.use(morgan('combined'));
   router.use([appRoutes.token], bodyParser.urlencoded({ extended: true }));
@@ -250,7 +238,7 @@ function startApp(issuer) {
     }
 
     var decoded = jwtDecode(tokens.access_token);
-    if (decoded.scp.indexOf('launch/patient') > -1) {
+    if ((decoded.scp != null) && (decoded.scp.indexOf('launch/patient') > -1)) {
       try {
         const response = await requestPromise({
           method: 'GET',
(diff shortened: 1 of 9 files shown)


oauth-proxy/tests/e2e.test.js
```diff
@@ -0,0 +1,229 @@
...
+
+  it('responds to the endpoints described in the OIDC metadata response', async () => {
+    // This test is making multiple requests. Theoretically it could be broken
+    // up, with each request being made in a separate test. That would make it
+    // much more difficult to use the metadata response to drive the requests
+    // for the subsequent requests.
+    const resp = await request({
+      methdod: 'get',
+      uri: 'http://localhost:9090/f/.well-known/openid-configuration',
+    });
```
dvogel: Oops. I'll fix this `methdod` typos.

oauth-proxy/tests/e2e.test.js
```diff
@@ -0,0 +1,229 @@
...
+
+afterAll(() => {
+  upstreamOAuthTestServer.stop();
+});
+
+const TEST_SERVER_PORT = 9090;
+const FAKE_CLIENT_APP_REDIRECT_URL = 'http://localhost:8080/oauth/redirect';
+const defaultTestingConfig = {
+  host: `http://localhost:${TEST_SERVER_PORT}`,
+  well_known_base_path: '/f',
```
kalilsn: Why `f`?

oauth-proxy/tests/e2e.test.js
```diff
@@ -0,0 +1,229 @@
...
+      resolveWithFullResponse: true,
+      method: 'get',
+      uri: parsedMeta.authorization_endpoint,
+      qs: {
+        client_id: 'clientId123',
+        state: 'abc123',
+        redirect_uri: 'http://localhost:8080/oauth/redirect',
+      },
+    });
+    expect(authorizeResp.statusCode).toEqual(302);
```
kalilsn: Is it worth checking the value of the location header here?

oauth-proxy/tests/e2e.test.js
```diff
@@ -0,0 +1,229 @@
...
+  });
+
+  it('responds to the SMART metadata endpoint', async () => {
+    const resp = await request({
+      method: 'get',
+      uri: 'http://localhost:9090/f/.well-known/smart-configuration.json',
+    });
+    const parsedMeta = JSON.parse(resp);
+    const expectedUrlPattern = `^${defaultTestingConfig.host}${defaultTestingConfig.well_known_base_path}.*`;
```
kalilsn: This is unused right now – were you intending to check the values in the object below?

oauth-proxy/package.json
```diff
@@ -24,11 +24,13 @@
...
     "jwt-decode": "^2.2.0",
     "morgan": "^1.9.1",
     "openid-client": "^2.4.5",
     "request": "^2.88.0",
-    "request-promise-native": "^1.0.5",
+    "request-promise-native": "^1.0.7",
     "yargs": "^12.0.5"
-  }
+  },
+  "devDependencies": {}
```
kalilsn: I think jest makes more sense as a dev dependency

oauth-proxy/tests/e2e.test.js
dvogel: It was `fakeBasePath` but that seemed too noisy in test failure messages. I'm fine using anything that's fairly short. Any suggestions?

oauth-proxy/tests/e2e.test.js
dvogel: Yep!

- [ ] Check authorize redirect destination URL.

oauth-proxy/tests/e2e.test.js
kalilsn: `test` maybe? Doesn't really matter, I was just curious 😆

oauth-proxy/tests/e2e.test.js
dvogel: I did. Good catch.

- [ ] Verify the format of the SMART endpoint URLs against the base URL of the test server.

oauth-proxy/package.json
dvogel: You're 100% right on that point. It's a mistake I commonly make with `npm install`

- [ ] Make jest a dev dependency.

oauth-proxy/tests/e2e.test.js
dvogel: I changed it to `testServer` because that corresponds to a bunch of other variable names for that same express app.

oauth-proxy/tests/e2e.test.js
dvogel: Uses the new `testServerBaseUrlPattern` now.

dvogel:
> Any reason not to add these to the Jenkinsfile now?

We should but I'm going to wait until a future PR for that because I need to refocus on adding the CORS support and testing it with my sample app changes.

oauth-proxy/tests/e2e.test.js
kalilsn: Just fyi, this one is checked off but the change isn't here

dvogel: Now it is :smile: Thanks for catching that.

oauth-proxy/tests/e2e.test.js
```diff
@@ -0,0 +1,232 @@
...
+      form: { grant_type: 'authorization_code', code: 'xzy789' },
+    });
+    // TODO: We should really call the token endpoint using the refresh_token
+    // grant type here. Right now the openid-client library makes this a little
+    // difficult. It automatically verifies the signature of the new access
+    // token. That's great, but doing full e2e testing would require making the
+    // upstream test server support constructing and signing proper JWTs. These
+    // tests should be enough to start breaking up the proxy app code into more
+    // easily testable parts and inject a fake openid client to side-step the
+    // signaure requirement.
```
mjlumetta: good note/comment

oauth-proxy/tests/e2e.test.js
mjlumetta: wait, maybe i haven't been reading this correctly, but shouldn't the value of that redirect match the `testServerBaseUrlPattern`?

dvogel: This is testing the initial request from the browser to the authorize endpoint. The redirect is sending them upstream to the authorization server to actually authenticate the user's identity. I'm going to re-order this test so that the authorize call comes before the token call, since that's closer to what would happen in a real login sequence.

oauth-proxy/index.js
```diff
@@ -291,25 +279,55 @@ function startApp(issuer) {
...
+  (async () => {
+    try {
+      const config = processArgs();
+      const issuer = await createIssuer(config);
+      startApp(config, issuer);
+    } catch (error) {
+      console.error(error);
+      process.exit(1);
+    }
+  })();
```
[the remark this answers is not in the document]
dvogel: I think that would be good but I was trying to make the absolute minimal diff I could in `index.js` to ensure that the tests were, as much as possible, testing what was there before I started adding tests.

oauth-proxy/tests/e2e.test.js
```diff
@@ -0,0 +1,232 @@
...
+  }
+}
+
+function buildFakeDynamoClient(fakeDynamoRecord) {
+  const dynamoClient = jest.genMockFromModule('../dynamo_client.js');
+  dynamoClient.saveToDynamo.mockImplementation((handle, state, key, value) => {
+    return new Promise((resolve, reject) => {
+      // It's unclear whether this should resolve with a full records or just
+      // the identity field but thus far it has been irrelevant to the
+      // functional testing of the oauth-proxy.
```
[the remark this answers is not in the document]
dvogel: I think we're on the same page about this. I want to significantly re-factor the dynamo_client.js module to be more self-contained, isolate the dynamodb handle more, and be more testable. For this PR though I wanted to make as few changes as I could to the existing code.

PR #43, opened 2019-06-12
Add CORS support to metadata endpoints.

dvogel:
This adds CORS support (both simple and pre-flight) to the metadata endpoints. This allows single page apps (including our react sample app) to auto-configure their OpenID Connect client.

Related to this ticket: https://github.com/department-of-veterans-affairs/vets-contrib/issues/1841

### DIFF
--- a/oauth-proxy/index.js
+++ b/oauth-proxy/index.js
@@ -1,4 +1,5 @@
 const express = require('express');
+const cors = require('cors');
 const { Issuer } = require('openid-client');
 const process = require('process');
 const { URL, URLSearchParams } = require('url');
@@ -80,7 +81,14 @@ function buildApp(config, issuer, oktaClient, dynamo, dynamoClient) {
   app.use(morgan('combined'));
   router.use([appRoutes.token], bodyParser.urlencoded({ extended: true }));
 
-  router.get('/.well-known/openid-configuration', (req, res) => {
+  const corsHandler = cors({
+    origin: true,
+    optionsSuccessStatus: 200,
+    preflightContinue: true,
+  });
+  router.options('/.well-known/*', corsHandler);
+
+  router.get('/.well-known/openid-configuration', corsHandler, (req, res) => {
     const baseMetadata = {...issuer.metadata, ...metadataRewrite }
     const filteredMetadata = openidMetadataWhitelist.reduce((meta, key) => {
       meta[key] = baseMetadata[key];
@@ -90,7 +98,7 @@ function buildApp(config, issuer, oktaClient, dynamo, dynamoClient) {
     res.json(filteredMetadata);
   });
 
-  router.get('/.well-known/smart-configuration.json', (req, res) => {
+  router.get('/.well-known/smart-configuration.json', corsHandler, (req, res) => {
     const baseMetadata = {...issuer.metadata, ...metadataRewrite }
     const filteredMetadata = smartMetadataWhitelist.reduce((meta, key) => {
       meta[key] = baseMetadata[key];
@@ -127,7 +135,7 @@ function buildApp(config, issuer, oktaClient, dynamo, dynamoClient) {
       res.redirect(`${document.redirect_uri.S}?${params.toString()}`)
     } catch (error) {
       console.error(error);
-      next(error); // This error is unrecoverable because we can't look up the original redirect.
+      return next(error); // This error is unrecoverable because we can't look up the original redirect.
     }
   });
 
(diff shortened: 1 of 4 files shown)


kalilsn:
I tried printing the response headers from the test request and I don't see any `Access-Control-Allow-Origin` header:
```
    { 'x-powered-by': 'Express',
      vary: 'Origin, Access-Control-Request-Headers',
      'access-control-allow-methods': 'GET,HEAD,PUT,PATCH,POST,DELETE',
      'access-control-allow-headers': '7bcf94b8007320984ba19b7e6c2d320333861ef0',
      allow: 'GET,HEAD',
      'content-type': 'text/html; charset=utf-8',
      'content-length': '8',
      etag: 'W/"8-ZRAf8oNBS3Bjb/SU2GYZCmbtmXg"',
      date: 'Wed, 12 Jun 2019 22:36:13 GMT',
      connection: 'close' }
```

dvogel:
> I tried printing the response headers from the test request and I don't see any `Access-Control-Allow-Origin` header:
> 
> ```
>     { 'x-powered-by': 'Express',
>       vary: 'Origin, Access-Control-Request-Headers',
>       'access-control-allow-methods': 'GET,HEAD,PUT,PATCH,POST,DELETE',
>       'access-control-allow-headers': '7bcf94b8007320984ba19b7e6c2d320333861ef0',
>       allow: 'GET,HEAD',
>       'content-type': 'text/html; charset=utf-8',
>       'content-length': '8',
>       etag: 'W/"8-ZRAf8oNBS3Bjb/SU2GYZCmbtmXg"',
>       date: 'Wed, 12 Jun 2019 22:36:13 GMT',
>       connection: 'close' }
> ```

The cors library is configured to allow cross-origin requests to the metadata endpoints from any origin. To do that, it reflects the value of the request's `origin` header in the response's `access-control-allow-origin` header. Since the test request didn't include an `origin` header, the response omitted the `access-control-allow-origin` header. Since the `request` library doesn't have the same inherent cross-origin concerns as a browser, it doesn't include an `origin` header by default. I've pushed a new commit to add one to the test request.

dvogel:
> Looks good! Thanks for adding that test!
> 
> > it reflects the value of the request's origin header in the response's access-control-allow-origin header
> 
> Just out of curiosity, is there a reason for doing that instead of just setting `Access-Control-Allow-Origin: *`?

There isn't one right now. Eventually we'll want to support CORS for `/token`. That will allow SPAs to use the authorization code grant type with the PKCE extension (I'm drafting a ticket for this now). Those requests will use `withCredentials` in order to include the `Authorization` header and `withCredential` requires an explicit origin (see ["Credentialed requests and wildcards"](https://developer.mozilla.org/en-US/docs/Web/HTTP/CORS))

oauth-proxy/index.js
```diff
@@ -127,7 +135,7 @@ function buildApp(config, issuer, oktaClient, dynamo, dynamoClient) {
       res.redirect(`${document.redirect_uri.S}?${params.toString()}`)
     } catch (error) {
       console.error(error);
-      next(error); // This error is unrecoverable because we can't look up the original redirect.
+      return next(error); // This error is unrecoverable because we can't look up the original redirect.
```
mjlumetta: what's the purpose of returning the `next` middleware function? i mean it's fine, but curious why you wanted to change them all (or unsure why it was necessary, per se)

oauth-proxy/tests/e2e.test.js
```diff
@@ -132,6 +134,23 @@ describe('OpenID Connect Conformance', () => {
     stopBackgroundServer();
   });
 
+  it('allows CORS on the OIDC metadta endpoint', async () => {
```
mjlumetta: typo in "metadata"

oauth-proxy/index.js
dvogel: The reason I changed all of them is because they are supposed to be error paths. The code below them is not supposed to run. What was happening was `next(error)` caused a response body to be sent to the browser. Then control flow returned from `next` and proceeded as if no error occurred. In the `authorize` case below, the next middleware was actually being called twice and finally failing with an internal error when the redirect tried to set a location header on a response object that had already sent the response body (and could thus no longer send another header).

Syntactically, it would be more appropriate to say:

```
next(error);
return;
```

Practically, I think the closure express wraps around the next middleware and passes as `next` always returns null so the `return next(error)` shows up in a lot of example code and it's become habit. I'm happy to change it but hopefully with the test suite in place we can start breaking up this code and defining this logic separate from the express control flow.

mjlumetta: nope, this is a good clarification! i think for some reason i was assuming that `res.end()` would terminate the control flow, but it's good to know that's not true.

PR #45, opened 2019-06-25
Upgrade express-prom-bundle to 5.1.5 and add it to oauth-proxy.

dvogel:
Adds express-prom-bundle to the oauth-proxy. Also upgrades it to 5.1.5 for the saml-proxy.

Related to: https://github.com/department-of-veterans-affairs/vets-contrib/issues/2407

### DIFF
--- a/oauth-proxy/index.js
+++ b/oauth-proxy/index.js
@@ -83,7 +83,7 @@ function buildApp(config, issuer, oktaClient, dynamo, dynamoClient) {
   app.use(promBundle({
     includeMethod: true,
     includePath: true,
-    customLables: {app: 'oauth_proxy'},
+    customLabels: {app: 'oauth_proxy'},
   }));
 
   router.use([appRoutes.token], bodyParser.urlencoded({ extended: true }));
--- a/oauth-proxy/index.js
+++ b/oauth-proxy/index.js
@@ -11,6 +11,7 @@ const { processArgs } = require('./cli')
 const okta = require('@okta/okta-sdk-nodejs');
 const morgan = require('morgan');
 const requestPromise = require('request-promise-native');
+const promBundle = require('express-prom-bundle');
 
 const appRoutes = {
   authorize: '/authorization',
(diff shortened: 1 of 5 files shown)


oauth-proxy/index.js
```diff
@@ -79,6 +80,12 @@ function buildApp(config, issuer, oktaClient, dynamo, dynamoClient) {
   const app = express();
   const router = new express.Router();
   app.use(morgan('combined'));
+  app.use(promBundle({
+    includeMethod: true,
+    includePath: true,
+    customLables: {app: 'oauth_proxy'},
+  }));
+
```
dvogel: The oauth-proxy seems to have fewer paths, so the normalization is a little less important here. Once we are collecting data we might find a need for normalization but that's hard to do by just looking at the code right now.

oauth-proxy/index.js
```diff
@@ -79,6 +80,12 @@ function buildApp(config, issuer, oktaClient, dynamo, dynamoClient) {
   const app = express();
   const router = new express.Router();
   app.use(morgan('combined'));
+  app.use(promBundle({
+    includeMethod: true,
+    includePath: true,
+    customLables: {app: 'oauth_proxy'},
```
kalilsn:
```suggestion
    customLabels: {app: 'oauth_proxy'},
```

kalilsn: Unless it's actually misspelled in express-prom-bundle :(

oauth-proxy/package.json
```diff
@@ -24,10 +24,12 @@
     "body-parser": "^1.18.3",
     "cors": "^2.8.5",
     "express": "^4.16.4",
+    "express-prom-bundle": "^5.1.5",
     "express-session": "^1.15.6",
     "jwt-decode": "^2.2.0",
     "morgan": "^1.9.1",
     "openid-client": "^2.4.5",
+    "prom-client": "^11.5.2",
```
kalilsn: Adding the explicit prom-client dependency seems like a good idea since we're going to use it for custom metrics. Perhaps we should do the same in the saml-proxy?

dvogel: Yeah, it's good to acknowledge that we have a direct dependency on it too. Although even if we didn't I think we'd have to do this because the newer versions of express-prom-bundle declare it as a peer dependency. I'll take a closer look at these package.json changes because I don't see a `peerDependencies` section.

PR #49, opened 2019-08-02
Increase headersTimeout from default 40s value

kalilsn:
Hopefully this resolves the 504 issue outlined in https://github.com/department-of-veterans-affairs/vets-contrib/issues/1094.

See my [comment on the issue](https://github.com/department-of-veterans-affairs/vets-contrib/issues/1094#issuecomment-517833070) for my justification of doing this without knowing whether it will help.

Also, 75s is an arbitrary choice, just keeping it the same as the keepalive setting. I don't really have any information to make a better choice, so speak up if you do!

### DIFF
--- a/oauth-proxy/index.js
+++ b/oauth-proxy/index.js
@@ -321,6 +321,7 @@ function startApp(config, issuer) {
   const env = app.get('env');
   const server = app.listen(config.port, () => console.log(`OAuth Proxy listening on port ${config.port} in ${env} mode!`));
   server.keepAliveTimeout = 75000;
+  server.headersTimeout = 75000;
   return null;
 }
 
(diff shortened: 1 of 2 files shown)


dvogel:
Seems like importing these values from the environment variables would make iteration a lot faster. I'd definitely approve a PR in that direction :)

kalilsn:
> Seems like importing these values from the environment variables would make iteration a lot faster. I'd definitely approve a PR in that direction :)

True, it would have removed the need for an additional build each time, which takes around 10 minutes. Unfortunately the deploys also take ~10 minutes, so it still wouldn't have been fast. I'm hopeful that we can make both of those processes a lot faster in the nearish future.

PR #50, opened 2019-08-08
Add extra logging to SAML Proxy

rtravitz:
This PR addresses [#2867](https://github.com/department-of-veterans-affairs/vets-contrib/issues/2867)

It does a few things:
1. Changes morgan's format to json so it's more easily consumable by Elasticsearch/Kibana later. Logs now look like this:
```
{"remote-address":"::1","time":"2019-08-08T15:37:37.803Z","method":"GET","url":"/samlproxy/idp/img/header-logo.png","status-code":"200","cont
ent-length":134182,"referrer":"http://localhost:7000/samlproxy/idp/saml/sso"}
```

2. Adds [winston](https://www.npmjs.com/package/winston) for logging. It's popular and powerful for creating re-usable formats and sending logs to a variety of end destinations. For example, it can add a meta tag to each log entry so that we can easily identify it came from the saml-proxy:
```
{"message":"[loadICN] - Calling MVI","level":"info","service":"saml-proxy"}
```

3. Adds logs around the MVI interaction. I need help thinking through what kind of session identifier we can use to tie these interactions together without exposing personal information that should remain secret.


saml-proxy/src/routes/logger.ts
```diff
@@ -0,0 +1,31 @@
...
+const jsonFormat = (tokens: TokenIndexer, req: Request, res: Response) => (
+  JSON.stringify({
+    'remote-address': tokens['remote-addr'](req, res),
+    'time': tokens['date'](req, res, 'iso'),
+    'method': tokens['method'](req, res),
+    'url': tokens['url'](req, res),
+    'status-code': tokens['status'](req, res),
+    'content-length': tokens['res'](req, res, 'content-length'),
+    'referrer': tokens['referrer'](req, res),
+  })
```
dvogel: If these fields of `tokens` are meant to always be present, can we use the accessor syntax, with an exception for the hyphenated option like the example does? https://github.com/expressjs/morgan#using-a-custom-format-function That's inline with the prefer-dot-notation eslint rule (https://eslint.org/docs/rules/dot-notation) and to my mind the values documented there way more heavily than consistent syntax on each row.

saml-proxy/src/routes/acsHandlers.ts
```diff
@@ -67,6 +69,7 @@ export const loadICN = async (req: IConfiguredRequest, res: Response, next: Next
       await req.vetsAPIClient.getVSOSearch(req.user.claims.firstName, req.user.claims.lastName);
       next();
     } catch (error) {
+      logger.error(`[loadICN] {session identifier} - Failed MVI lookup and VSO search: ${error}`)
```
dvogel: I think the `catch` block that starts on line 63 /65 would also be useful to log.

This logging will rely on the `toString` function of the error object. Do we know that the errors raised by the callees here will have a useful `toString` function? If my memory serves me right, some of these might be coming from the request library (see this comment: https://github.com/department-of-veterans-affairs/vets-saml-proxy/pull/50/files#diff-c5d6e220feba0aaf2f111560a96b457cR12)

rtravitz: I can log that, too. We can probably get an interesting chart out of how many VSOs are going through the OAuth flow based on it.

Checked what comes from the `toString`. This is an example of a failed lookup: `{"message":"[loadICN] {session identifier} - Failed MVI lookup and VSO search: StatusCodeError: 404 - {\"errors\":[{\"detail\":\"Representative not found\"}]}","level":"error","service":"saml-proxy"}`.

rtravitz:
I think I took care of the requested changes @dvogel. I also attached a UUID that we can use to identify sessions. Logs now look like this with slightly more structure:
1. Arriving from Okta
```
{"service":"saml-proxy","action":"parseSamlRequest","result":"success","session":"a997871f-2853-4e60-b39e-8eed7e16a35d","level":"info","messa
ge":"User arrived from Okta. Rendering IDP login template."}
```

2. Failing the MVI lookup and returning the error after I shut of my local `vets-api` server
```
{"service":"saml-proxy","action":"loadICN","result":"failure","session":"232c9458-7ccd-4803-ad53-d945d087dfef","level":"error","message":"Failed MVI lookup and VSO search: RequestError: Error: connect ECONNREFUSED 127.0.0.1:5000"}
```

dvogel:
Nice improvements. Let's get these deployed and tested! :smile:

rtravitz:
There were some unrelated Jenkins issues with saml-proxy deployments yesterday and today. Got one through this afternoon to staging, and logs are coming through as expected to the `dvp-staging-saml-proxy` group 👍

PR #53, opened 2019-08-19
Add VA Internal CA Certs

wunderhund:
This is in preparation for setting up new certs on the vets-api ELBs which are signed by the VA's internal CA certs rather than using the public cert there. This PR adds these VA CA certs to the saml-proxy deployment's docker containers so that it will trust SSL connections to internal VA https sites (including, before long, api.va.gov, dev-api.va.gov, staging-api.va.gov, etc)

Before you ask, I was unable to figure out a good way to put all the certs in one location and pull them into each docker container, because the containers can only copy in files within their context. Unfortunately, just putting all the certs into each container's directory seems like the best and most straightforward way to get this done.


kalilsn:
I'm realizing now that perhaps the claims_v0 service won't trigger those alerts? If that's the case, then go ahead and dismiss my review and merge!

PR #56, opened 2019-11-12
Add sentry to oauth/saml proxies

jasalisbury:
https://github.com/department-of-veterans-affairs/vets-contrib/issues/2838

Adds sentry reporting to oauth/saml proxies. Secrets are filtered out in `beforeSend`. I found the fields that need to be filtered by watching requests, let me know if you know of any more that need filtered. We could start logging all requests the proxies see in `dev` to look for more but that means we could be logging secrets in `dev`.

There is a bit of duplication between the saml and oauth proxies. Let me know if you can think of a good way of sharing some of the Sentry code across "projects".

~~Leaving as a draft until the Sentry projects get setup. I'm not a Sentry admin so I cannot create the projects myself.~~ I'm now a Sentry Admin (thanks Kalil) and have successfully created the Sentry projects.

### DIFF
--- a/oauth-proxy/index.js
+++ b/oauth-proxy/index.js
@@ -12,6 +12,7 @@ const okta = require('@okta/okta-sdk-nodejs');
 const morgan = require('morgan');
 const requestPromise = require('request-promise-native');
 const promBundle = require('express-prom-bundle');
+const Sentry = require('@sentry/node');
 
 const appRoutes = {
   authorize: '/authorization',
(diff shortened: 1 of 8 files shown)


oauth-proxy/index.js
```diff
@@ -72,7 +73,34 @@ function buildMetadataRewriteTable(config, appRoutes) {
   };
 }
 
+function filterProperty(object, property) {
+  if (property in object) {
+    object[property] = '[Filtered]';
```
jasalisbury: I replace the property with `[Filtered]` to match filtering that Sentry does. I could see making this string distinct from the string Sentry uses to help distinguish where the filtering happens.

oauth-proxy/index.js
```diff
@@ -72,7 +73,34 @@ function buildMetadataRewriteTable(config, appRoutes) {
...
+    Sentry.init({
+      dsn: process.env.SENTRY_DSN,
+      environment: process.env.SENTRY_ENVIRONMENT,
+      beforeSend(event) {
+        filterProperty(event.request, 'cookies');
+        filterProperty(event.request.headers, 'cookie');
+        filterProperty(event.request.headers, 'authorization');
+        return event;
+      },
+      shouldHandleError(error) {
```
jasalisbury: I could see use handling some 4xx errors along with all 5xx errors.

oauth-proxy/index.js
```diff
@@ -81,7 +109,12 @@ function buildApp(config, issuer, oktaClient, dynamo, dynamoClient) {
   const router = new express.Router();
   // Express needs to know it is being ran behind a trusted proxy. Setting 'trust proxy' to true does a few things
   // but notably sets req.ip = 'X-Forwarded-for'. See http://expressjs.com/en/guide/behind-proxies.html
-  app.set('trust proxy', true)
+  app.set('trust proxy', true);
+  if (useSentry) {
+    app.use(Sentry.Handlers.requestHandler({
+      user: false,
```
jasalisbury: We have no obvious user id field in these requests. I guess we _might_ be able to use `client_id` but I haven't given that too much thought.

oauth-proxy/index.js
dvogel: Sure let's add 4xx here too. We can dial it back later if it's too noisy.

oauth-proxy/index.js
```diff
@@ -72,7 +73,36 @@ function buildMetadataRewriteTable(config, appRoutes) {
...
 }
 
+function filterProperty(object, property) {
+  if (property in object) {
+    object[property] = '[Filtered]';
+  }
+}
+
 function buildApp(config, issuer, oktaClient, dynamo, dynamoClient) {
+  const useSentry = typeof config.sentry_dsn !== 'undefined' && typeof config.sentry_environment !== 'undefined';
```
mjlumetta: could we check if the value is undefined directly? since undefined is also the value

jasalisbury: so `config.sentry_dsn !== undefined`?

mjlumetta: yup, that's what i was thinking

PR #63, opened 2019-11-15
Log relay state

jasalisbury:
https://github.com/department-of-veterans-affairs/vets-contrib/issues/3617

Relay state is logged in three places, in order of where they are encountered in the login flow:

 * Coming from Okta to the saml proxy (the first call the saml proxy sees in the login flow)
 * Coming from the IDP back to the saml proxy (after the user has logged in)
 * Going from the saml proxy back to Okta (this log takes place in the same call as the second bullet point). There is potential that this relay state is different than what it was coming from the IDP.

There is a session Id that is logged along side the relay state to help match up requests. At first I was concerned with the security of logging this session ID but it is already logged all over the saml-proxy so this at least doesn't increase any risk (might want to make sure we want that session id logged though). All the information is split into fields in a JSON object to help querying against it. An iso8601 datetime was added as well.

I increased a session's max age to help make sure we keep the same session for the same login process. Some users have to enter an MFA token or may be slower logging in. I know we can zoom through the login process in dev but we don't really need to enter any credentials. Honestly it _seems_ reasonable to me to increase it further (maybe to five minutes?) but one minute to two minutes is probably sufficient for now.

### Additional Notes

I didn't find anything from Okta that could be used to matchup the SAML request coming into the saml proxy to the saml response coming into the saml proxy from the IDP, which is why I used the session id. I even looked in the `SamlRequest` and `SamlResponse` objects (let me know if I missed something)


saml-proxy/src/routes/acsHandlers.ts
```diff
@@ -114,6 +115,17 @@ export const serializeAssertions = (req: IConfiguredRequest, res: Response, next
...
+      datetime: new Date().toISOString(),
+      relayState: authOptions.RelayState
+    };
+
+    logger.info(`Relay state to Okta (from session): ${authOptions.RelayState}`, logObj);
+  } else {
+    logRelayState(req, logger, 'to Okta');
   }
   authOptions.authnContextClassRef = req.user.authnContext.authnMethod;
   samlp.auth(authOptions)(req, res, next);
```
jasalisbury: `RelayState` in the request might be overwritten by authOptions, see the [docs](https://github.com/auth0/node-samlp#login-authentication-flow). This is why there are log lines for both paths through the branch on line 116, we need to know and log if we use the `RelayState` from the session or the request.

As an aside, the docs say the default for `RelayState` is `req.query.RelayState`. I've never seen `RelayState` in the query, it has always been on the body. In my testing I always use the session version. I'm trying to track down when the `sessions` `RelayState` isn't used.

jasalisbury: I thought about this over the weekend a bit and now log `req.query.RelayState` if used. If it is used I believe it will be `null` and potentially the source of the login issues we are experiencing.

PR #66, opened 2019-11-19
Add metrics for user lookups

jasalisbury:
https://github.com/department-of-veterans-affairs/vets-contrib/issues/2393

Add prometheus metrics for MVI and VSO lookups. Using different buckets than the default (.005, .01, .025, .05, .1, .25, .5, 1, 2.5, 5, 10) but let me know if I should adjust from what I currently have.

### DIFF
--- a/saml-proxy/src/metrics.ts
+++ b/saml-proxy/src/metrics.ts
@@ -1,0 +1,54 @@
+import client from 'prom-client';
+
+const defaultLabels = { app: 'saml_proxy' };
+client.register.setDefaultLabels(defaultLabels);
+const latencyBuckets = [0.1, 0.25, 0.5, 1, 2, 5, 10, 15];
+const bucketLabels = ['status_code'];
+
+export const MVILookupBucket = new client.Histogram({
+  name: 'mvi_lookup_bucket',
+  help: 'durtation histogram of duration of MVI lookup requests',
+  buckets: latencyBuckets,
+  labelNames: bucketLabels,
+});
+
+export const MVIAttempt = new client.Counter({
+  name: 'mvi_lookup_attempt',
+  help: 'counter of number of lookup requests sent to MVI',
+});
+
+export const MVIFailure = new client.Counter({
+  name: 'mvi_lookup_failure',
+  help: 'counter of number of lookup request failures from MVI',
+});
+
+export const VSOLookupBucket = new client.Histogram({
+  name: 'vso_lookup_bucket',
+  help: 'durtation histogram of duration of VSO lookup requests',
+  buckets: latencyBuckets,
+  labelNames: bucketLabels,
+});
+
+export const VSOAttempt = new client.Counter({
+  name: 'vso_lookup_attempt',
+  help: 'counter of number of lookup requests sent to VSO',
+});
+
+export const VSOFailure = new client.Counter({
+  name: 'vso_lookup_failure',
+  help: 'counter of number of lookup request failures from VSO',
+});
+
+export async function requestWithMetrics(histogram: client.Histogram, attempt: client.Counter, failure: client.Counter, promise: Promise<any>) {
+  const timer = histogram.startTimer();
+  attempt.inc();
+  try {
+    var res = await promise;
+    timer({status_code: '200'});
+    return res;
+  } catch(err) {
+    failure.inc();
+    timer({status_code: err.statusCode || 'unknown'});
+    throw err;
+  }
+}
--- a/saml-proxy/src/metrics.ts
+++ b/saml-proxy/src/metrics.ts
@@ -2,13 +2,16 @@ import client from 'prom-client';
 
 const defaultLabels = { app: 'saml_proxy' };
 client.register.setDefaultLabels(defaultLabels);
-const latencyBuckets = [0.1, 0.25, 0.5, 1, 2, 5, 10, 15];
+// Default buckets are [.005, .01, .025, .05, .1, .25, .5, 1, 2.5, 5, 10]. There are too many buckets
+// on the low end and the max bucket is too high as well. Anything above 5 is too high. To help focus
+// the histogram on latencies we care about we only use a subset of the default
+const latencyBucketsSecs = [0.05, 0.1, 0.25, 0.5, 1, 2.5, 5];
 const bucketLabels = ['status_code'];
 
 export const MVILookupBucket = new client.Histogram({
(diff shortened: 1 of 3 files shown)


saml-proxy/src/metrics.ts
```diff
@@ -0,0 +1,54 @@
+import client from 'prom-client';
+
+const defaultLabels = { app: 'saml_proxy' };
+client.register.setDefaultLabels(defaultLabels);
+const latencyBuckets = [0.1, 0.25, 0.5, 1, 2, 5, 10, 15];
+const bucketLabels = ['status_code'];
```
dvogel: Given the values, I assume these are in seconds? I'd like to see the variable name remove the ambiguity. E.g. `latencyBucketsSecs`.

Separately, a comment explaining the rationale for these bucket values would be nice. I'd like to see an additional lower bucket. Maybe `.05` -- not unreasonably to try to push the best case down under 50ms. On the other hand, I don't think there's much value in differentiating between 10 and 15 seconds because both are unacceptably long, even as outliers. I would like to see an explicit timeout (maybe configurable via env var) for the outbound HTTP requests and then have the last histogram be roughly half that timeout. IOW 15 makes sense if our timeout is 30s but that's probably much, much to high for a timeout here. About 10s is the absolute max we should tolerate, in which case the top histogram bucket should be 5+.

saml-proxy/src/metrics.ts
```diff
@@ -0,0 +1,57 @@
...
+const defaultLabels = { app: 'saml_proxy' };
+client.register.setDefaultLabels(defaultLabels);
+// Default buckets are [.005, .01, .025, .05, .1, .25, .5, 1, 2.5, 5, 10]. There are too many buckets
+// on the low end and the max bucket is too high as well. Anything above 5 is too high. To help focus
+// the histogram on latencies we care about we only use a subset of the default
+const latencyBucketsSecs = [0.05, 0.1, 0.25, 0.5, 1, 2.5, 5];
+const bucketLabels = ['status_code'];
+
+export const MVILookupBucket = new client.Histogram({
+  name: 'mvi_lookup_requset_duration_seconds_bucket',
```
kalilsn:
```suggestion
  name: 'mvi_lookup_request_duration_seconds_bucket',
```

kalilsn: `s` and `e` in request were transposed

saml-proxy/src/metrics.ts
```diff
@@ -0,0 +1,57 @@
...
+  help: 'counter of number of lookup requests sent to MVI',
+});
+
+export const MVIFailure = new client.Counter({
+  name: 'mvi_lookup_failure',
+  help: 'counter of number of lookup request failures from MVI',
+});
+
+export const VSOLookupBucket = new client.Histogram({
+  name: 'vso_lookup_requset_duration_seconds_bucket',
```
kalilsn:
```suggestion
  name: 'vso_lookup_request_duration_seconds_bucket',
```

saml-proxy/src/metrics.ts
```diff
@@ -0,0 +1,57 @@
...
+  help: 'durtation histogram of duration of VSO lookup requests labeled with: app, status_code',
```
kalilsn:
```suggestion
  help: 'histogram of duration of VSO lookup requests labeled with: app, status_code',
```

saml-proxy/src/metrics.ts
```diff
@@ -0,0 +1,57 @@
...
+client.register.setDefaultLabels(defaultLabels);
+// Default buckets are [.005, .01, .025, .05, .1, .25, .5, 1, 2.5, 5, 10]. There are too many buckets
+// on the low end and the max bucket is too high as well. Anything above 5 is too high. To help focus
+// the histogram on latencies we care about we only use a subset of the default
+const latencyBucketsSecs = [0.05, 0.1, 0.25, 0.5, 1, 2.5, 5];
+const bucketLabels = ['status_code'];
+
+export const MVILookupBucket = new client.Histogram({
+  name: 'mvi_lookup_requset_duration_seconds_bucket',
+  help: 'durtation histogram of duration of MVI lookup requests labeled with: app, status_code',
```
kalilsn:
```suggestion
  help: 'histogram of duration of MVI lookup requests labeled with: app, status_code',
```

kalilsn: I know this help message is shown at the /metrics endpoint, but do you know if it's exposed in the prometheus UI anywhere? If so I would argue for making these a little more verbose

jasalisbury: I don't think it is exposed anywhere

saml-proxy/src/metrics.ts
```diff
@@ -0,0 +1,57 @@
...
+  buckets: latencyBucketsSecs,
+  labelNames: bucketLabels,
+});
+
+export const MVIAttempt = new client.Counter({
+  name: 'mvi_lookup_attempt',
+  help: 'counter of number of lookup requests sent to MVI',
+});
+
+export const MVIFailure = new client.Counter({
```
mjlumetta: since these client metric classes are intended to be consumed together in `requestWithMetrics`, can they be rolled into one aggregate object with a histogram, an attempt counter, and a failure counter? if we're going to want to be more flexible in the future, then maybe it's not helpful, but if we do that it can simplify the `requestWithMetrics` signature a little.

saml-proxy/src/metrics.ts
```diff
@@ -0,0 +1,57 @@
...
+  name: 'vso_lookup_attempt',
+  help: 'counter of number of lookup requests sent to VSO',
+});
+
+export const VSOFailure = new client.Counter({
+  name: 'vso_lookup_failure',
+  help: 'counter of number of lookup request failures from VSO',
+});
+
+export async function requestWithMetrics(histogram: client.Histogram, attempt: client.Counter, failure: client.Counter, promiseFunc: () => Promise<any>) {
```
mjlumetta: what would you think about moving this to the `acsHandlers` file? i think that since it's responsible for making the HTTP request, it would be nice to have it next to the route handlers. but it doesn't matter that much, either is fine. either way, no changes to the body, looks good.

dvogel:
I think all the changes look good but I think we should add some tests. E.g. I think a unit test to verify that `requesWithMetrics` actually calls the `promiseFunction` would be good.

PR #68, opened 2019-11-21
Add prometheus counters to track oauth login flow

jasalisbury:
https://github.com/department-of-veterans-affairs/vets-contrib/issues/3697

This will add one counter to track number of users that have started the OAuth login process. It also will add a counter to track the number of users that have finished the OAuth login process. Together we should be able to get an idea about how many users are failing to complete the login process.

### DIFF
--- a/oauth-proxy/metrics.js
+++ b/oauth-proxy/metrics.js
@@ -1,0 +1,19 @@
+const client = require('prom-client');
+
+const defaultLabels = { app: 'oauth_proxy' };
+client.register.setDefaultLabels(defaultLabels);
+
+const loginBegin = new client.Counter({
+  name: 'oauth_login_begin',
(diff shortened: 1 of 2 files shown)


oauth-proxy/metrics.js
```diff
@@ -0,0 +1,19 @@
...
+
+const loginBegin = new client.Counter({
+  name: 'oauth_login_begin',
+  help: 'counter of number of times the OAuth login process has begun',
+});
+
+const loginEnd = new client.Counter({
+  name: 'oauth_login_end',
+  help: 'counter of number of times the OAuth login process has ended',
+});
```
dvogel: Can we rename this to `oauth_proxy_login_end` (ditto for `_begin`)? I could see VSP ending up with other metrics referring to "oauth" in the future. Best to be specific about what component is being measured IMO.

PR #69, opened 2019-11-29
Convert OAuth and SAML proxy logs to JSON and add timestamp

mjlumetta:
https://github.com/department-of-veterans-affairs/vets-contrib/issues/3568

* Use Winston logger in OAuth proxy
* Create Winston JSON logger 
* Use `timestamp` format in all Winston loggers on top of JSON (do we prefer "timestamp" or "time" for this property name - is one more consistent with other components or have conflicts, or is it sort of a 6 on one hand, half dozen on the other type of decision?)
* Some updates to improve running OAuth proxy locally

### DIFF
--- a/oauth-proxy/cli.js
+++ b/oauth-proxy/cli.js
@@ -71,23 +71,7 @@ function processArgs() {
     .argv;
 }
 
-function processSchemaArgs() {
-  return yargs
-    .usage("DynamoDB schema migration to set up the OAuth Proxy")
-    .options({
-      local: {
-        boolean: true,
-        default: false,
-        description: 'Flag to use localhost instead of Docker container hostname',
-        required: false,
-      }
-    })
-    .wrap(yargs.terminalWidth())
-    .argv;
-}
-
 module.exports = {
   processArgs,
-  processSchemaArgs,
-}
+};
 
--- a/oauth-proxy/cli.js
+++ b/oauth-proxy/cli.js
@@ -71,7 +71,23 @@ function processArgs() {
     .argv;
 }
 
+function processSchemaArgs() {
+  return yargs
+    .usage("DynamoDB schema migration to set up the OAuth Proxy")
+    .options({
+      local: {
+        boolean: true,
+        default: false,
+        description: 'Flag to use localhost instead of Docker container hostname',
+        required: false,
+      }
+    })
+    .wrap(yargs.terminalWidth())
+    .argv;
+}
+
 module.exports = {
   processArgs,
+  processSchemaArgs,
 }
 
--- a/oauth-proxy/index.js
+++ b/oauth-proxy/index.js
@@ -2,7 +2,7 @@ const express = require('express');
 const cors = require('cors');
 const { Issuer } = require('openid-client');
 const process = require('process');
-const { URL, URLSearchParams } = require('url');
+const { URLSearchParams } = require('url');
 const bodyParser = require('body-parser');
 const request = require('request');
 const jwtDecode = require('jwt-decode');
@@ -13,6 +13,7 @@ const morgan = require('morgan');
 const requestPromise = require('request-promise-native');
 const promBundle = require('express-prom-bundle');
 const Sentry = require('@sentry/node');
+const logger = require('./logger');
 
 const appRoutes = {
   authorize: '/authorization',
@@ -171,7 +172,7 @@ function buildApp(config, issuer, oktaClient, dynamo, dynamoClient) {
       try {
         await dynamoClient.saveToDynamo(dynamo, state, "code", req.query.code);
       } catch (error) {
-        console.error(error);
+        logger.error(error);
       }
     }
     try {
@@ -179,7 +180,7 @@ function buildApp(config, issuer, oktaClient, dynamo, dynamoClient) {
       const params = new URLSearchParams(req.query);
       res.redirect(`${document.redirect_uri.S}?${params.toString()}`)
     } catch (error) {
-      console.error(error);
+      logger.error(error);
       return next(error); // This error is unrecoverable because we can't look up the original redirect.
     }
   });
@@ -197,7 +198,7 @@ function buildApp(config, issuer, oktaClient, dynamo, dynamoClient) {
         return res.redirect(`${client_redirect}?${errorParams.toString()}`);
       }
     } catch (error) {
-      console.error(error);
+      logger.error(error);
       // This error is unrecoverable because we would be unable to verify
       // that we are redirecting to a whitelisted client url
       return next(error);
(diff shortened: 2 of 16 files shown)


saml-proxy/src/routes/acsHandlers.ts
```diff
@@ -5,7 +5,7 @@ import { IConfiguredRequest } from './types';
 import { NextFunction, Response } from "express";
 import assignIn from 'lodash.assignin';
 import samlp from "samlp"; import * as url from "url";
-import logger from './logger';
+import logger from '../logger';
```
dvogel: Now that the logger module exports an object with two keys, does this need to be changed to destructure that object like this?

```
import { logger } from '../logger';
```

saml-proxy/src/routes/acsHandlers.ts
```diff
@@ -76,7 +76,7 @@ export const loadICN = async (req: IConfiguredRequest, res: Response, next: Next
     next();
   } catch (error) {
     const mviError = error;
-    logger.info(`Failed MVI lookup; will try VSO search: ${error}`, { session, action, result: 'failure' });
+    logger.warn(`Failed MVI lookup; will try VSO search`, { error, session, action, result: 'failure' });
```
dvogel: Now that `error` will be a key inside the logged object, perhaps we should use `mviError`? I think we can get rid of the `mviError = error` assignment and use `mviError` as the name of the caught exception.

oauth-proxy/cli.js
```diff
@@ -71,7 +71,23 @@ function processArgs() {
...
+        boolean: true,
+        default: false,
+        description: 'Flag to use localhost instead of Docker container hostname',
+        required: false,
+      }
+    })
+    .wrap(yargs.terminalWidth())
+    .argv;
+}
+
```
dvogel: I like this option because it recognizes that we need a friendlier way to handle our dynamodb dependency. Having this part of the same module as the yargs processor for the main app feels awkward to me since the schema migrations are still executed separately from the rest of the app. I think there's a strong argument to be made for having a single app, making the main app responsible for handling it's own schema migration. That seems like too much to include in this PR though so for now could we move this into `dynamo_schema.ts`?

saml-proxy/src/routes/acsHandlers.ts
mjlumetta: nope, it exports those things but still has a default export. so if i wanted to import the whole module, piece by piece:

```
import logger, { loggingMiddleware, sassLogger } from '../logger';
```

i think it makes sense to have the general-purpose logger as the default export.

oauth-proxy/cli.js
mjlumetta: yup, that sounds good to me. in general, i think we should invest in improving local development for these components as we work with them. i added this because i was trying to run the OAuth proxy locally rather than in docker and couldn't get it to work (this part did, but requests from the running proxy said the table didn't exist?). anyway, used docker in the end because it wasn't worth the time, but i think that making it easier to work with locally and documenting how to do so will pay significant dividends in terms of speeding up development on the two proxies in the long term. to that end, i also have a PR for the client app that i'm going to open. anyway, all of your concrete suggestions sound good to me.

saml-proxy/src/routes/acsHandlers.ts
mjlumetta: nope, it exports those things but still has a default export. so if i wanted to import the whole module, piece by piece:

```
import logger, { loggingMiddleware, sassLogger } from '../logger';
```

i think it makes sense to have the general-purpose logger as the default export.

oauth-proxy/cli.js
dvogel: > i think we should invest in improving local development for these components as we work with them.

I 100% agree. If you have ideas for how to improve these, let's pursue them. If they are small (like this) feel free to include them in other work. If they are large, feel free to write them up as tickets of their own.

dvogel:
> LGTM! Is Winston's timestamp field an actual unix timestamp (12345...) or a date (2019-12-02T....)? As you alluded I don't think it matters that much but I'd want the field to be called `timestamp` if it is a unix timestamp or `time`/`datetime` if it is a date format (iso8601). Matching any logs that already hit elastic search/kibana in name and format is most important.

This is a great question! I think we should use iso8601 timestamps. A unix timestamp is ambiguous re: timezone. The time datatypes in elasticsearch understand iso8601, so using that format doesn't preclude sorting or range queries (which seems to be the usual motivation for using unix timestamps).

saml-proxy/src/logger.ts
```diff
@@ -2,9 +2,23 @@ import morgan, { TokenIndexer, Options } from 'morgan';
 import { format, createLogger, transports } from 'winston';
 import { Request, Response } from 'express';
 
+const logFormat = format.combine(
+  format.timestamp({ alias: "time" }),
+  format.json({ 
+    space: 2,
+    replacer: (key: string, value: {}) => {
+      if (typeof value === "object" && key === "" && "timestamp" in value) {
+        delete value["timestamp"];
```
mjlumetta: the `alias` of the timestamp format is addition to "timestamp", not instead of it. this is to prevent duplicating that and potentially causing confusion in the future

mjlumetta:
@jasalisbury @dvogel for posterity, the timestamp format uses the ISO 8601 format: https://github.com/winstonjs/logform#timestamp

oauth-proxy/index.js
```diff
@@ -267,7 +268,7 @@ function buildApp(config, issuer, oktaClient, dynamo, dynamoClient) {
           {...req.body, redirect_uri }
         );
       } catch (error) {
-        console.log(error.response);
+        logger.error(error.response, error);
```
dvogel: Why log both of these? It seems like logging `error` will include `error.response`, so logging both would be duplicating content. Is there something I'm missing?

saml-proxy/src/logger.ts
dvogel: @mjlumetta can you add a comment to the code explaining that so we can't forget when we look at this again in a year?

oauth-proxy/index.js
mjlumetta: yeah, i think you're right that this is a mistake, but it's actually probably something that should change everywhere. the first param to `logger.error()` (or info, debug, warn, etc) is the message itself, and the second param is the meta properties object that gets mapped to the resulting JSON log. the two required properties for each info object - the object that gets passed to the logger/transport - are the level, which is set by the `severity` param in `logger.log()` or automatically when a specific level method is used. so i think these all need to be `logger.error("descriptive message", error)`. i was trying to use `error.response` as the descriptive message but that doesn't really make any more sense. relevant winston documentation: https://github.com/winstonjs/winston#streams-objectmode-and-info-objects

saml-proxy/src/logger.ts
mjlumetta: absolutely. by the way - i think this would be an easy opportunity to make an open-source contribution to the `logform` package, if we feel so inclined. it's annoying to me that this is necessary

oauth-proxy/index.js
jasalisbury: I actually change this line in https://github.com/department-of-veterans-affairs/vets-saml-proxy/pull/71. `error.response` can be undefined so probably shouldn't be used for the message.

PR #70, opened 2019-11-29
Run oauth proxy tests on jenkins

kalilsn:
https://github.com/department-of-veterans-affairs/vets-contrib/issues/3671

More tests to come, but for now we should at least make sure the existing oauth proxy tests get run on each commit.

### DIFF
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -26,7 +26,7 @@ pipeline {
       steps {
         sh '''
           cd /opt/app
-          npm run test:ci'
+          npm run test:ci
         '''
       }
       post {
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -14,7 +14,7 @@ pipeline {
       }
     }
 
-    stage('Run tests') {
+    stage('Run saml-proxy tests') {
       agent {
         dockerfile {
           args "--entrypoint='' -u 0:0"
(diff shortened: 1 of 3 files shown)


kalilsn:
>Do we need to install packages before running tests?

We already install packages when we [build the Dockerfile](https://github.com/department-of-veterans-affairs/vets-saml-proxy/blob/kalilsn/run-oauth-proxy-tests/oauth-proxy/Dockerfile#L10). Adding `npm install` to the Jenkinsfile would cause us to install them twice for each run.

However, we do need to `cd /opt/app` (the `WORKDIR` set in the dockerfile) to take advantage of those already installed packages, since Jenkins overrides the workdir setting at runtime.

>Why not use npm run-script test?

No reason, other than I didn't notice that was defined. I'll switch!

>Do we need to generate and gather test results in an xml file (like is done for the saml-proxy tests?)
I don't think so. We could do that, but it hides test failures behind a second click that I've seen lots of people miss in the past. I think just printing failures to the console is the best solution for the way we use Jenkins. If we were relying on Jenkins to calculate some kind of metrics based on the number of tests passed/failed, then I think it would be worth switching to an xml report.

kalilsn:
>1. what does Jenkins override the directory to? haven't run into that, would like to learn 😄

It got set to `/home/jenkins/workspace/xy_kalilsn_run-oauth-proxy-tests` on my last run. If you search for that string [in the log](http://jenkins.vfs.va.gov/blue/rest/organizations/jenkins/pipelines/testing/pipelines/vets-saml-proxy/branches/kalilsn%252Frun-oauth-proxy-tests/runs/11/log/?start=0) you can see the full command. Let me know if you think there's a better way to do this! I'll go ahead and make sure we run the SAML and OAuth proxy tests the same way before merging this.

>2. in the SAML proxy tests, what does withEnv(['CI=true']) do? couldn't find anything with a quick grep. if nothing, can it be removed here?

I think you're right that it doesn't do anything. I assumed the `CI` var meant something to `jest` but I can't find any evidence of that. Nice catch!

PR #71, opened 2019-12-02
Safely extract status code from error object

jasalisbury:
https://github.com/department-of-veterans-affairs/vets-contrib/issues/3727

If the error object does not have a response or status code we default to 500. 500 is a number and not a string to make sure we return the same type from `statusCodeFromError` (https://nodejs.org/api/http.html#http_message_statuscode).


oauth-proxy/index.js
```diff
@@ -345,6 +347,13 @@ function buildApp(config, issuer, oktaClient, dynamo, dynamoClient) {
   return app;
 }
 
+function statusCodeFromError(error) {
```
mjlumetta: would you consider creating a new `utils.js` file and moving this there? i don't love that the OAuth proxy is almost entirely defined in this one file; i'm not 100% sure, but i think that may be a legacy of Ed or whoever writing the minimal version of this component when it was initially created? regardless, would love to see us break things out a little more as we grow this component. not super important given the size of this function but i think it would set a nice precedent

jasalisbury: yeah that is probably a smart move... Also will make it easier to have unit tests for this function.

dvogel: The single file overly simplistic aspect of this stems from the Sample Application section of the authorization docs: https://developer.va.gov/explore/health/docs/authorization#sample-application

oauth-proxy/index.js
```diff
@@ -271,8 +273,9 @@ function buildApp(config, issuer, oktaClient, dynamo, dynamoClient) {
           {...req.body, redirect_uri }
         );
       } catch (error) {
-        console.log(error.response);
-        res.status(error.response.statusCode).json({
+        console.log(error);
```
kalilsn: Now that https://github.com/department-of-veterans-affairs/vets-saml-proxy/pull/69 is merged, can you convert this to use the winston logger?

PR #79, opened 2020-01-09
Refactor token validation to enable e2e test coverage for launch/pati…

dvogel:
This PR seeks to cover the `launch/patient` path in `tokenHandler`. This path was previously uncovered in the e2e token test because the mocked upstream oauth server was returning an access token that lacked the `launch/patient` scope.

Once that path was covered, I learned that the overly broad `catch` blocks were swallowing a `ReferenceError` [in my previous PR to extract tokenHandler to it's own file](https://github.com/department-of-veterans-affairs/vets-saml-proxy/pull/78/files). I've wrapped each `catch` block in a call to the new `stdExcFilter` to help catch these reference errors as more of `index.js` is extracted into other modules.

Finally, in order to provide a mocked token validation response I extracted the upstream call to it's own function. In addition to mocking the response, the extracted method is also paired with a configuration method that allows us to put a line in the sand re: refactoring. The `tokenHandler` method already has a reference to the config object but it would be nice to pass all configurable behaviors into `tokenHandler` already configured and get rid of that parameter to `tokenHandler`.


oauth-proxy/oauthHandlers/tokenHandler.js
```diff
@@ -79,23 +89,21 @@ const tokenHandler = async (config, redirect_uri, logger, issuer, dynamo, dynamo
...
-        json: true,
-        headers: {
-          apiKey: config.validate_apiKey,
-          authorization: `Bearer ${tokens.access_token}`,
-        }
-      });
-      const patient = response.data.attributes.va_identifiers.icn;
+      const validation_result = await validateToken(tokens.access_token);
+      const patient = validation_result.va_identifiers.icn;
+      console.log({...tokens, patient, state});
```
jasalisbury: Did you mean for this log to stay in?

dvogel: Nope :)

PR #80, opened 2020-01-09
Use expires_in instead of expires_at

dvogel:
When the proxy requests the grant from the upstream oauth server it receives a `TokenSet` object ([docs here](https://github.com/panva/node-openid-client/blob/master/docs/README.md#tokenset)). The library translates the `expires_in` field from the upstream server to an `expires_at` field, translating from a relative timestamp to an absolute timestamp. The proxy was serializing this object directly, causing the client application to see a non-conformant token response that lacked an `expires_in` field. This PR fixes that by making the proxy fully responsible for the shape of the response to the client application.

### DIFF
--- a/oauth-proxy/oauthHandlers/tokenResponse.js
+++ b/oauth-proxy/oauthHandlers/tokenResponse.js
@@ -1,0 +1,65 @@
+// Returns an object suitable for the response body of an OAuth 2.0 token
+// response. The input is a TokenSet object from the openid-client library. See
+// documentation here:
+// https://github.com/panva/node-openid-client/blob/master/docs/README.md#tokenset
+//
+// The returned object will always have these keys:
+//   * access_token
+//   * expires_in
+//   * refresh_token
+//   * scope
+// If the token set is for an OIDC request it will also have:
+//   * id_token
+const translateTokenSet = (token_set) => {
+  // Example successful response from RFC 6749
+  // {
+  //   "access_token":"2YotnFZFEjr1zCsicMWpAA",
+  //   "token_type":"example",
+  //   "expires_in":3600,
+  //   "refresh_token":"tGzv3JOkF0XG5Qx2TlKWIA",
+  //   "example_parameter":"example_value"
+  // }
+
+  // Example successful response from OpenID Connect Core 1.0
+  // {
+  //  "access_token": "SlAV32hkKG",
+  //  "token_type": "Bearer",
+  //  "refresh_token": "8xLOxBtZp8",
+  //  "expires_in": 3600,
+  //  "id_token": "eyJhbGciOiJSUzI1NiIsImtpZCI6IjFlOWdkazcifQ.ewogImlzc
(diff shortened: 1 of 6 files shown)


oauth-proxy/oauthHandlers/tokenResponse.js
```diff
@@ -0,0 +1,60 @@
...
+    if (token_set[field] != null) {
+      responseAccum[field] = token_set[field];
+    }
+  };
+  copy_field('access_token');
+  copy_field('id_token');
+  copy_field('refresh_token');
+  copy_field('token_type');
+  copy_field('scope');
+  debugger
```
jasalisbury:
```suggestion
```

oauth-proxy/tests/oauthHandlers.test.js
```diff
@@ -0,0 +1,48 @@
...
+  });
+
+  it('translates absolute timestamps to relative timestamps', async () => {
+    const abs_oauth_set = new TokenSet({
+      access_token: 'oauth.is.cool',
+      refresh_token: 'refresh.later',
+      expires_in: 7200,
+    });
+    const now_sec = Math.floor(Date.now() / 1000);
+    expect(abs_oauth_set.expires_at - now_sec).toBeGreaterThanOrEqual(7200);
```
jasalisbury: I've had success in the past using [timecop](https://github.com/travisjeffery/timecop) in Ruby for tests like this. Quick search for a similar library in javascript shows [timekeeper](https://github.com/vesln/timekeeper). Might be nice to get familiar with a javascript tool that can ease the development of time depended tests in case we need to write more in the future.

oauth-proxy/oauthHandlers/tokenResponse.js
dvogel: Nice catch!

oauth-proxy/oauthHandlers/tokenResponse.js
```diff
@@ -0,0 +1,59 @@
...
+    if (token_set[field] != null) {
+      responseAccum[field] = token_set[field];
+    }
+  };
+  copy_field('access_token');
+  copy_field('id_token');
+  copy_field('refresh_token');
+  copy_field('token_type');
+  copy_field('scope');
+  responseAccum['expires_in'] = token_set.expires_in;
```
mjlumetta: can you add a comment clarifying why this is? it's pretty unintuitive as is, both in the context of the PR and i would guess outside of it too. looking at the source for the OpenID Client, it looks like it's because `expires_in` is defined with a getter/setter based on `expires_at`. but that's not clear from this code or their docs, unless i'm missing something. 

similarly, is there any reason not to use `copy_field` for this as well?

dvogel: Sure, I can add a comment. Using `copy_field` is possible. I originally thought `TokenSet.expires_in` was a method rather than a property. That's why I used different syntax to begin with.

mjlumetta: cool, thanks! looks good to me otherwise, just had trouble reading this line

PR #82, opened 2020-01-10
Add saml proxy functional tests

jasalisbury:
https://github.com/department-of-veterans-affairs/vets-contrib/issues/1352

Adds functional testing for the saml-proxy. There is a test that makes sure relay state stays the same and then four tests for each identity provider, only implemented for idme in this PR. There are three tests that deal with finding (or not finding) the user in MVI or VSO and another test to make sure users without the proper LOA are handled correctly. Also there is a small refactor for the saml-proxy so the `vets-api` client is mockable.

Most tests require some XML parsing. If the test request makes it all the way through successfully the `SAMLResponse` and `RelayState` are embedded in html. The `SAMLResponse` may require additional xml parsing to pull out specific attributes.

Cert/key pairs were generated by me locally for the specific purpose of these tests.

At this point it should be easy to add additional tests so if anyone thinks of anything missing let me know!


saml-proxy/test/e2e.test.js
```diff
@@ -0,0 +1,258 @@
...
+    uri: `http://localhost:${PORT}/samlproxy/sp/saml/sso`,
+    form: {
+      SAMLResponse: samlResponse,
+      RelayState: state,
+    }
+  };
+
+  return request(reqOpts);
+}
+
```
jasalisbury: I could see these parser functions being factored into their own class(es) in the future. Doesn't feel warranted at this point but I could be convinced otherwise!

saml-proxy/test/e2e.test.js
```diff
@@ -0,0 +1,258 @@
...
+    });
+
+    it('properly handles user not found in mvi or vso', async () => {
+      const requestSamlResponse = await buildSamlResponse('user', '3');
+      const response = await ssoRequest(requestSamlResponse);
+      expect(responseResultType(response)).toEqual(USER_NOT_FOUND);
+    });
+  });
+
+  // describe.skip('dslogin', () => {
```
jasalisbury: I'm planning on implementing `dslogin` and `mhv` tests in another PR to follow this one

saml-proxy/test/e2e.test.js
```diff
@@ -0,0 +1,258 @@
+require('jest');
+
+import request from 'request-promise-native';
+import { getSamlResponse } from 'samlp';
+import { DOMParser } from 'xmldom';
+
+import { buildBackgroundServerModule } from '../../oauth-proxy/tests/backgroundServer';
```
jasalisbury: Note this is importing code written in the `oauth-proxy` which means this will break if we ever split the `oauth-proxy` and `saml-proxy` into separate repos. I see two ways forward (let me know if you see any other!)

 1. We add a note here so if we ever split the projects we have some direction
 2. We just copy the code from the `oauth-proxy` over to the `saml-proxy` right now and just have some duplicate code

kalilsn: So the tests are failing right now because they run in a docker container that only has the source code for a single one of these applications. What about pulling this out into a `common` directory that we explicitly add to both containers for shared modules?

jasalisbury: Not sure this is the way to go after looking at it a bit. We'd have to change the docker build context to include the folder above the `saml-proxy` and `oauth-proxy` folders to include the common folder, which doesn't seem great. Maybe there is a better way to do it? Let me know if you think so and we can pair on it because I don't see a great way to do it... I think I'm leaning towards just duplicating the code in the `saml-proxy` folder.

kalilsn: I'm not sure I agree that it's a big problem to include the parent folder in the build context. The only addition should be the source code from the `oauth-proxy`, which should be negligible. And since we're not actually adding those files to the image it shouldn't have any effect on the actual image size. Any other files beyond the source code that would be added by increasing the build context should probably be excluded in a .dockerignore anyways.

That said, I haven't actually implemented this and there may be considerations I'm missing, so I'm happy to pair with you on this, or just accept whatever decision you make!

jasalisbury: maybe having the build context be a folder up seemed like a bigger deal then it actually is... I'll take a crack at it and see what it looks like!

kalilsn: We should add `.git` to the `.dockerignore` regardless

saml-proxy/test/SAMLResponse.example.xml
```diff
@@ -0,0 +1,68 @@
...
+      </saml:Attribute>
+      <saml:Attribute Name="uuid" NameFormat="urn:oasis:names:tc:SAML:2.0:attrname-format:basic">
+        <saml:AttributeValue xsi:type="xs:string">ae9ff5f4e4b741389904087d94cd19b2</saml:AttributeValue>
+      </saml:Attribute>
+      <saml:Attribute Name="level_of_assurance" NameFormat="urn:oasis:names:tc:SAML:2.0:attrname-format:basic">
+        <saml:AttributeValue xsi:type="xs:string">3</saml:AttributeValue>
+      </saml:Attribute>
+    </saml:AttributeStatement>
+  </saml:Assertion>
+</samlp:Response>
```
dvogel: Since we don't publish the `vets.gov.user` test users anymore, maybe it would be best to use `va.api.user+idme` test users in these e2e tests?

saml-proxy/test/e2e.test.js
```diff
@@ -0,0 +1,258 @@
...
+      const responseSamlResponse = atob(SAMLResponseFromHtml(response.body));
+      const icn = assertionValueFromSAMLResponse(responseSamlResponse, 'icn');
+      expect(icn).toBeUndefined();
+    });
+
+    it('properly handles user not found in mvi or vso', async () => {
+      const requestSamlResponse = await buildSamlResponse('user', '3');
+      const response = await ssoRequest(requestSamlResponse);
+      expect(responseResultType(response)).toEqual(USER_NOT_FOUND);
+    });
```
dvogel: The word "properly" feels underspecific and overused. I think something like like "provides downstream user-not-found error when user not found in mvi or vso" lets the reader know what is proper. For the test above something like "relies on vso lookup when the user is not known to mvi" might be an improvement?

saml-proxy/test/e2e.test.js
dvogel: Why do we have to parse this HTML? I thought the SAMLResponse was POST'd back to the saml-proxy.

jasalisbury: The `saml-proxy` responds to the `sso` request with [this template](https://github.com/auth0/node-samlp/blob/master/templates/form.ejs). The `sso` endpoint doesn't directly redirect but renders HTML/javascript that automatically redirects when rendered by the user's user-agent. We could setup these tests with puppeteer (or something similar) to have the SAMLResponse get POST'd somewhere we can grab for the tests. Note if we go this route we'd have to fire up another server to capture the POST and make the result available in the test. Since we already have to do some XML parsing for the SAMLResponse it seemed reasonable to just pull out the SAMLResponse from the HTML/javascript returned from the `sso` endpoint. Parsing the SAMLReponse from HTML seems less complicated and easier to do, but open to other suggestions!

kalilsn: So the saml-proxy serves a form to the user which is automatically submitted and POSTs the SAMLResponse to the saml-proxy? Why do we need that step at all?

jasalisbury: I'm starting to think I just didn't setup the server correctly... looking into it.

jasalisbury: Screenshot from 2020-01-13 13-34-11
Just went through the login process locally and it appears to serves a form that is automatically submitted.

As an aside I was previously confused as to why I wasn't seeing the form in the network requests in google chrome. Turns out the response is lost inside google's developer tools once you navigate away from the page the request occurred on even when using "preserve logs" (which is why the screenshot uses firefox). Something to file away in the back of the mind when troubleshooting in the future.

kalilsn: I'm sure you're right about how it works! I'm honestly just confused about why the saml proxy works that way. It just seems unnecessarily convoluted to have our backend serve a form that is automatically submitted back to itself. I'm sure there's something I'm missing that makes this much more reasonable than it seems right now. Maybe you can explain it to me on a call at some point this week

kalilsn: And sorry for derailing this conversation. I think leaving out the browser and just parsing the HTML here is a very reasonable approach for this kind of test.

jasalisbury: I needed to confirm that is how it works because as you say it seems overly complicated! I'm still trying to track down why it works this way. I assume I am missing something as well...

dvogel: I just hopped on a call to review this with @jasalisbury. I believe the reason for the auto-submitted form is due to inconsistent behavior in browers wrt POST redirects (see the difference between 301/302 vs 307/308).

Josh and I identified where the template rendering behavior is configured for the IDP side of the proxy. We could technically stub this out with something more easily testable. Since that would diminish the test's claim to being e2e, I vote we stick with the awkward HTML parsing bits.

For my part, I was thrown off more than I should have been by some small inconsistencies in the template we have vs the one provided by the auth0 library. That was compounded by simple boolean bug that Josh and I found in the `responseHandler` call.

saml-proxy/test/e2e.test.js
```diff
@@ -65,8 +65,15 @@ function ssoRequest(samlResponse, state = 'state') {
...
-// for the html document being parsed
+// These are the HTML parsers. The SSO endpoint renders HTML/javascript that automatically
+// submits a form of the SAMLResponse and RelayState back to Okta. It doesn't simply redirect
+// because it would need to do a 307 redirect to stay a POST. The user would need to "approve"
+// this redirect (see https://www.w3.org/Protocols/rfc2616/rfc2616-sec10.html). The `samlp`
+// library we use allows the form template to be specified via config. We define the template
+// to use in in the function `responseHandler` in  `src/IDPConfig`. See `src/views/samlresponse.hbs`
+// for the html we are parsing. We decided to get the SAMLResponse and RelayState by parsing the
+// resulting HTML instead of mocking out the response handler. Parsing the HTML keeps the
+// functionality of these tests closer to what users actually experience.
```
dvogel: :+1:

kalilsn: Ah, this clears things up. I was confused because the form's `action` points back to the saml proxy in the tests, whereas it actually points to okta when live.

PR #85, opened 2020-01-10
Make tokenHandler call next() in all paths.

dvogel:
This should fix [bug #4029](https://github.com/department-of-veterans-affairs/vets-contrib/issues/4029). Essentially the `tokenHandler` error paths were failing to `return`, so we tried to continue processing the request after the error. In addition to that, both error and success paths were not properly calling the `next` expressjs callback. I found this while trying to write the new tests for `tokenHandler` in `oauthHandlers.test.js`. Yay testing! :)

### DIFF
--- a/oauth-proxy/tests/oauthHandlers.test.js
+++ b/oauth-proxy/tests/oauthHandlers.test.js
@@ -2,10 +2,16 @@
 
 require('jest');
 
-const { TokenSet } = require('openid-client');
+const MockExpressRequest = require('mock-express-request');
+const MockExpressResponse = require('mock-express-response');
+const { Client, Issuer, TokenSet } = require('openid-client');
+const { RequestError } = require('request-promise-native/errors');
 const timekeeper = require('timekeeper');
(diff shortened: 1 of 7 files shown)


oauth-proxy/tests/oauthHandlers.test.js
```diff
@@ -52,3 +58,213 @@ describe('translateTokenSet', () => {
...
+
+class FakeIssuer {
+  constructor(client) {
+    this.Client = class FakeInlineClient {
+      constructor(_) {
+        return client;
+      }
+    };
+  }
+}
```
dvogel: TIL you can generate inline es6 classes that share a name. I'm not sure whether I want to laugh, jump for joy, or cry. Maybe all three? :smile:

kalilsn: I'm struggling to think of a legitimate use for that "feature" but it seems like a great way to annoy your collaborators 😈

PR #87, opened 2020-01-14
Add functional test cases for mhv and dslogin

jasalisbury:
https://github.com/department-of-veterans-affairs/vets-contrib/issues/1352

Major changes from the [last PR](https://github.com/department-of-veterans-affairs/vets-saml-proxy/pull/82) are the mock vets-api client is easier to control and the user in the SAMLResponse hitting the `sso` endpoint can be from `idme`, `mhv` or `dslogin`. Note the loop in the e2e test that runs the same tests for all three identity providers.

### DIFF
--- a/saml-proxy/test/e2e.test.js
+++ b/saml-proxy/test/e2e.test.js
@@ -203,10 +203,10 @@ describe('Logins for idp', () => {
         expect(icn).toEqual('123');
       });
 
-      it('looks up user from vso if the lookup from mvi fails', async () => {
+      it('treats the user as a VSO if the lookup from mvi fails', async () => {
         const requestSamlResponse = await buildSamlResponse(idp, '3');
         vetsApiClient.findUserInMVI = false;
-        vetsApiClient.findUserInVSO = true;
+        vetsApiClient.userIsVSO = true;
         const response = await ssoRequest(requestSamlResponse);
 
         expect(responseResultType(response)).toEqual(SAML_RESPONSE);
@@ -216,10 +216,10 @@ describe('Logins for idp', () => {
         expect(icn).toBeUndefined();
       });
 
-      it('returns a user not found page when the user is not found in mvi or vso', async () => {
+      it('returns a user not found page when the user is not found in mvi or is not a VSO', async () => {
         const requestSamlResponse = await buildSamlResponse(idp, '3');
         vetsApiClient.findUserInMVI = false;
-        vetsApiClient.findUserInVSO = false;
+        vetsApiClient.userIsVSO = false;
         const response = await ssoRequest(requestSamlResponse);
         expect(responseResultType(response)).toEqual(USER_NOT_FOUND);
       });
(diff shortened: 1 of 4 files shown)


saml-proxy/test/testUsers.js
```diff
@@ -0,0 +1,73 @@
...
+      level_of_assurance: '0',
+      dslogon_assurance: loa,
+      dslogon_mname: 'E',
+      multifactor: 'true',
+      uuid: 'ae9ff5f4e4b741389904087d94cd19b2'
+    }
+  }
+}
+
+function idmUser(level_of_assurance) {
```
kalilsn:
```suggestion
function idmeUser(level_of_assurance) {
```

saml-proxy/test/testUsers.js
```diff
@@ -0,0 +1,73 @@
...
+      multifactor: 'true',
+      uuid: 'ae9ff5f4e4b741389904087d94cd19b2'
+    }
+  };
+}
+
+export function getUser(type, level_of_assurance) {
+  switch(type) {
+    case IDME_USER:
+      return idmUser(level_of_assurance);
```
kalilsn:
```suggestion
      return idmeUser(level_of_assurance);
```

saml-proxy/test/e2e.test.js
```diff
@@ -191,75 +182,47 @@ describe('Logins for idp', () => {
...
+
+        const responseSamlResponse = atob(SAMLResponseFromHtml(response.body));
+        const icn = assertionValueFromSAMLResponse(responseSamlResponse, 'icn');
+        expect(icn).toEqual('123');
+      });
+
+      it('looks up user from vso if the lookup from mvi fails', async () => {
+        const requestSamlResponse = await buildSamlResponse(idp, '3');
+        vetsApiClient.findUserInMVI = false;
+        vetsApiClient.findUserInVSO = true;
```
kalilsn: Could we rename these things a little bit? The way the test/method on vetsApiClient are named right now it sounds like VSO is an alternative service to MVI. I think the test name should be more like `'treats user as a VSO if not found in MVI'`

kalilsn: And maybe the `findUserInVSO` property isn't necessary at all? Instead of looking for that to be true could you test for `!foundUserInMVI`?

jasalisbury: > Could we rename these things a little bit? The way the test/method on vetsApiClient are named right now it sounds like VSO is an alternative service to MVI. I think the test name should be more like `'treats user as a VSO if not found in MVI'`

It sounded like that because that is what I thought. Thanks for jogging my brain so I can realize they aren't separate services! I don't know why it didn't connect for me that VSO is a Veteran Service Organization.

> And maybe the findUserInVSO property isn't necessary at all? Instead of looking for that to be true could you test for !foundUserInMVI?

Unfortunately don't think we can just use `!findUserInMvi` because we want to test the case where the user isn't in MVI and the user isn't a VSO. I'll rename the `findUserInVSO` attribute.

PR #88, opened 2020-01-15
Update to node 12

jasalisbury:
https://github.com/department-of-veterans-affairs/vets-contrib/issues/3671

Verified login works locally. Combined cert is still being used, node prints a warning when it cannot load certs from `NODE_EXTRA_CA_CERTS`. I don't see that error when I run the oauth or saml proxy. I can see the combined cert in the correct location in the docker containers and the `NODE_EXTRA_CA_CERTS` variable set (also in the docker containers) to point to the combined cert.


jasalisbury:
> All the changes here look legit for moving to v13. However, moving from v8 to v13 (even vs odd) takes us from the stable release track to the unstable release track. Does this gain us anything significant over staying on the stable release track but upgrading to v12?

I don't think v13 will give us anything we can't get in v12, it's probably a better decision to upgrade to v12.

PR #89, opened 2020-01-31
Fix oauth refresh timeouts

kalilsn:
https://github.com/department-of-veterans-affairs/vets-contrib/issues/4132

See https://lighthouseva.slack.com/archives/CMT4MFPS6/p1580154255007800 and https://lighthouseva.slack.com/archives/CGG5E8CUU/p1580220425011000 for additional context.

This PR fixes a small programming error (using `document.state.S` when `document` might be `undefined`) and ensures that we will actually send a 500 response when encountering an unexpected error. Additionally, it modifies the refresh handler so that errors retrieving the `state` param from DynamoDB will only prevent the `state` from being included in the response, rather than sending an error to the user.

### DIFF
--- a/oauth-proxy/index.js
+++ b/oauth-proxy/index.js
@@ -150,28 +150,31 @@ function buildApp(config, issuer, oktaClient, dynamo, dynamoClient, validateToke
     res.json(filteredMetadata);
   });
 
-  router.get(appRoutes.jwks, async (req, res) => {
+  router.get(appRoutes.jwks, (req, res) => {
     req.pipe(request(issuer.metadata.jwks_uri)).pipe(res)
   });
 
-  router.get(appRoutes.userinfo, async (req, res) => {
+  router.get(appRoutes.userinfo, (req, res) => {
     req.pipe(request(issuer.metadata.userinfo_endpoint)).pipe(res)
   });
 
-  router.post(appRoutes.introspection, async (req, res) => {
+  router.post(appRoutes.introspection, (req, res) => {
     req.pipe(request(issuer.metadata.introspection_endpoint)).pipe(res)
   });
 
   router.get(appRoutes.redirect, async (req, res, next) => {
-    await oauthHandlers.redirectHandler(logger, dynamo, dynamoClient, req, res, next);
+    await oauthHandlers.redirectHandler(logger, dynamo, dynamoClient, req, res, next)
+      .catch(next)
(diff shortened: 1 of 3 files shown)


oauth-proxy/index.js
```diff
@@ -150,49 +150,54 @@ function buildApp(config, issuer, oktaClient, dynamo, dynamoClient, validateToke
...
     req.pipe(request(issuer.metadata.jwks_uri)).pipe(res)
   });
 
-  router.get(appRoutes.userinfo, async (req, res) => {
+  router.get(appRoutes.userinfo, (req, res) => {
     req.pipe(request(issuer.metadata.userinfo_endpoint)).pipe(res)
   });
 
-  router.post(appRoutes.introspection, async (req, res) => {
+  router.post(appRoutes.introspection, (req, res) => {
```
kalilsn: This function and those above have been transformed to synchronous handlers to take advantage of the automatic handling provided by express

oauth-proxy/index.js
```diff
@@ -150,49 +150,54 @@ function buildApp(config, issuer, oktaClient, dynamo, dynamoClient, validateToke
...
     req.pipe(request(issuer.metadata.introspection_endpoint)).pipe(res)
   });
 
   router.get(appRoutes.redirect, async (req, res, next) => {
-    await oauthHandlers.redirectHandler(logger, dynamo, dynamoClient, req, res, next);
+    await oauthHandlers.redirectHandler(logger, dynamo, dynamoClient, req, res, next)
+      .catch(next)
```
kalilsn: These async handlers all have to explicitly define an error callback. We can simply use the `next` function as the error callback here since the only argument it needs is the error itself. This should be equivalent to:

```javascript
try {
	await oauthHandlers.redirectHandler(logger, dynamo, dynamoClient, req, res, next)
} catch(err) {
	next(err)
}
```

kalilsn:
> You mentioned a conversation with Drew about why testing this satisfactorily would be difficult. Can you share a bit more about what the problems are?

@nfasulo I was just referring to [this comment](https://github.com/department-of-veterans-affairs/vets-saml-proxy/blob/master/oauth-proxy/tests/e2e.test.js#L199-L206), which captures the difficulty I was having in writing an e2e test for this component. It's not insurmountable, just more difficult than the simple thing I was anticipating!

PR #91, opened 2020-02-26
Piv 216 additional oauth proxy timing metrics

esubocz-va:
Add timing metrics through oauth-proxy flow.
related story: https://vasdvp.atlassian.net/browse/PIV-216

### DIFF
--- a/oauth-proxy/oauthHandlers/tokenHandler.js
+++ b/oauth-proxy/oauthHandlers/tokenHandler.js
@@ -4,13 +4,13 @@ const requestPromise = require('request-promise-native');
 const { rethrowIfRuntimeError, statusCodeFromError } = require('../utils');
 const { translateTokenSet } = require('./tokenResponse');
 
-const { tokenHandlerGauge } = require('../metrics');
+const { refreshGauge } = require('../metrics');
+const { dynamodbGauge } = require('../metrics');
+const { grantGauge } = require('../metrics');
+
 
 const tokenHandler = async (config, redirect_uri, logger, issuer, dynamo, dynamoClient, validateToken, req, res, next) => {
   let client_id, client_secret;
-  tokenHandlerGauge.setToCurrentTime();
-  const end = tokenHandlerGauge.startTimer();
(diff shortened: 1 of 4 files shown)


beaugrantham-va:
I think we should make the following changes.

1. We should use `gauge.set(...)` instead of `startTimer()`. The concern here is thread safety. Since multiple requests could be in-flight, and `gauge` is a shared object, there is the possibility of capturing inaccurate values. We'll have to do the timing on our own (using [`process.hrtime.bigint()`](https://nodejs.org/api/process.html#process_process_hrtime_bigint)).
2. Let's drop the gauge for the database. I'm not sure it adds value. To really add value we'd want to measure each type of read and write to see where performance is an issue, and we'd want to do it across the entire application. Too much overhead for now.
3. Move the token validation gauge into `tokenValidation.js` so any client will reap the benefits.
4. The refresh gauge should be renamed to something like `oktaRefreshTokenGauge` to emphasize Okta, and should also capture a value when the request fails.

oauth-proxy/oauthHandlers/tokenHandler.js
```diff
@@ -32,11 +34,13 @@ const tokenHandler = async (config, redirect_uri, logger, issuer, dynamo, dynamo
 
   let tokens, state;
   if (req.body.grant_type === 'refresh_token') {
+    const oktaTokenRefreshStart = process.hrtime.bigint();
     try {
       tokens = await client.refresh(req.body.refresh_token);
+      const oktaTokenRefreshEnd = process.hrtime.bigint();
+      oktaTokenRefreshGauge.set(Number(oktaTokenRefreshEnd - oktaTokenRefreshStart)/1000000000);
```
beaugrantham-va: We should also time failed requests (inside the catch block).

Also since we're doing this in multiple places, it's probably worth adding a function inside `utils.js` or somewhere that we can just call... something like `stopTimer(oktaTokenRefreshGauge, oktaTokenRefreshStart);`.

PR #92, opened 2020-03-06
Task/piv 243 replace depricated library

esubocz-va:
related story: https://vasdvp.atlassian.net/secure/RapidBoard.jspa?rapidView=130&projectKey=PIV&modal=detail&selectedIssue=PIV-243

### DIFF
--- a/oauth-proxy/package.json
+++ b/oauth-proxy/package.json
@@ -22,6 +22,7 @@
     "@okta/okta-sdk-nodejs": "^1.2.0",
     "@sentry/node": "^5.7.1",
     "aws-sdk": "^2.372.0",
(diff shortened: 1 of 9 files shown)


oauth-proxy/package.json
```diff
@@ -32,7 +33,6 @@
     "openid-client": "^2.4.5",
     "prom-client": "^11.5.2",
     "request": "^2.88.0",
```
beaugrantham-va: We should also replace `request`, since it's the reason that `request-promise-native` was deprecated.

beaugrantham-va: As discussed, we'll open a new task to replace `request`.

PR #94, opened 2020-03-21
PIV-260 - Better Basic Auth Handling

beaugrantham-va:
https://vasdvp.atlassian.net/browse/PIV-260

Adds better handling for malformed basic authorization headers.


oauth-proxy/metrics.js
```diff
@@ -23,9 +23,15 @@ const validationGauge = new client.Gauge({
   help: 'metric for timing of validation flow'
 });
 
+function stopTimer(gauge, start) {
```
beaugrantham-va: This function was just moved from `utils.js`.

PR #95, opened 2020-03-26
PIV-280 - Configurable OIDC timeout

beaugrantham-va:
https://vasdvp.atlassian.net/browse/PIV-280

Adds an optional config to override the default 1500ms timeout for OIDC requests.

See https://github.com/department-of-veterans-affairs/devops/pull/6589 for related config changes (15s timeout in non-prod environments).

### DIFF
--- a/oauth-proxy/cli.js
+++ b/oauth-proxy/cli.js
@@ -20,7 +20,7 @@ function processArgs() {
       },
       upstream_issuer_timeout_ms: {
         description: 'Optional timeout (ms) for upstream requests',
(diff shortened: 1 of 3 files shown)


beaugrantham-va:
Timeouts can be simulated using a local oauth-proxy by sending Okta traffic to a bad host.

```
$ docker exec -it -u 0 oauth-proxy_oauth-proxy_1 sh
/opt/app # echo "10.255.255.1 [okta fqdn]" >> /etc/hosts
```

PR #97, opened 2020-03-27
PIV-260 reapply basic auth parsing

beaugrantham-va:
Reapply PIV-260.

The original branch had one import for `stopTimer` that was not updated after moving the function. Was not caught in testing because it requires a specific scope to be requested.

### DIFF
--- a/oauth-proxy/tokenValidation.js
+++ b/oauth-proxy/tokenValidation.js
@@ -1,6 +1,5 @@
 const process = require('process');
-const { validationGauge } = require('./metrics');
-const { stopTimer } = require('./utils');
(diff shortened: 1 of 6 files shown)


oauth-proxy/tokenValidation.js
```diff
@@ -1,6 +1,5 @@
 const process = require('process');
-const { validationGauge } = require('./metrics');
-const { stopTimer } = require('./utils');
+const { validationGauge, stopTimer } = require('./metrics');
```
beaugrantham-va: This was the change that was missed on the original PR.

PR #99, opened 2020-04-02
PIV-277 - PKCE Auth Flow

beaugrantham-va:
https://vasdvp.atlassian.net/browse/PIV-277

Add support for PKCE auth flow.

This can be tested with a local oauth-proxy using the following client.  
 * https://github.com/beaugrantham-va/vets-api-clients/tree/PIV-277-sample-pkce-client/samples/oauth_pkce_cli
 * NOTE: The `client_id` must be configured for PKCE. At least one test application is registered if you search for "PKCE".
 * NOTE: Add `"enable_pkce_authorization_flow": true` in `dev-config.json`.
```
npm i
./pkce-cli \
  --client_id 00000000000000000000 \
  --auth_server http://localhost:7100 \
  --scopes "openid profile email offline_access launch/patient" \
  --redirect_uri http://localhost:8080/redirect 
```

These changes are backwards compatible with the existing auth flow (using a `client_secret`) with one item of note.
* A missing `client_secret` previously threw a 401 immediately without proxying the traffic to the auth server. Because we now support PKCE clients, these requests do get proxied through and we rely on the auth server to throw a 401 and populate the error message for invalid client credentials.

### DIFF
--- a/oauth-proxy/cli.js
+++ b/oauth-proxy/cli.js
@@ -70,6 +70,11 @@ function processArgs() {
         required: false,
         string: true
       },
+      enable_pkce_authorization_flow: {
+        description: "Enable PKCE authorization flow?",
+        required: false,
+        default: false,
+      }
     })
     .wrap(yargs.terminalWidth())
     .argv;
--- a/oauth-proxy/cli.js
+++ b/oauth-proxy/cli.js
@@ -52,7 +52,7 @@ function processArgs() {
       validate_endpoint: {
         description: 'va.gov token validation endpoint',
         required: true,
-        default: 'https://dev-api.va.gov/internal/auth/v0/validation'
+        default: 'https://sandbox-api.va.gov/internal/auth/v0/validation'
       },
       validate_apiKey: {
         description: 'apiKey permitting access to validate endpoint',
--- a/oauth-proxy/oauthHandlers/tokenHandler.js
+++ b/oauth-proxy/oauthHandlers/tokenHandler.js
@@ -1,16 +1,15 @@
 const jwtDecode = require('jwt-decode');
 const process = require('process');
 
-const { rethrowIfRuntimeError, statusCodeFromError, stopTimer } = require('../utils');
+const { rethrowIfRuntimeError, statusCodeFromError, parseBasicAuth } = require('../utils');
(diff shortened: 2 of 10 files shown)


oauth-proxy/cli.js
```diff
@@ -70,6 +70,11 @@ function processArgs() {
         required: false,
         string: true
       },
+      enable_pkce_authorization_flow: {
+        description: "Enable PKCE authorization flow?",
+        required: false,
+        default: false,
```
beaugrantham-va, line 76: This toggle will allow us to have sufficient runtime and testing in dev/sandbox before we promote to production. An accompanying config PR will follow this.

oauth-proxy/oauthHandlers/tokenHandler.js
```diff
@@ -6,14 +6,25 @@ const { translateTokenSet } = require('./tokenResponse');
...
-    ([ client_id, client_secret ] = [ basicAuth.username, basicAuth.password ]);
+    clientMetadata.client_id = basicAuth.username;
+    clientMetadata.client_secret = basicAuth.password;
   } else if (req.body.client_id && req.body.client_secret) {
-    ({ client_id, client_secret } = req.body);
+    clientMetadata.client_id = req.body.client_id;
+    clientMetadata.client_secret = req.body.client_secret;
     delete req.body.client_id;
     delete req.body.client_secret;
+  } else if (config.enable_pkce_authorization_flow && req.body.client_id) {
```
kalilsn, line 24: If you wanted to preserve the quick failure for a client sending an empty client_secret but not intending to use PKCE, you could check for the presence of the `code_verifier` param here. I'm not sure that really matters though!

beaugrantham-va: This token handler is used for both `authorization_code` and `refresh_token`. The
`code_verifier` param is only present for the `authorization_code` grant. We'd have to break up the two or make the conditional slightly more specific.

kalilsn: Yeah, it would require more refactoring. Breaking this endpoint up doesn't sound like a terrible idea to me from a readability perspective, but I don't have any issue with this as it is!

PR #101, opened 2020-04-14
Change Entrypoint to not use entrypoint script

dginther:
Right now since we are running on two separate platforms, the entrypoint script does not function correctly. 

This bypasses it.

I will create a separate Dockerfile for the fargate process that will not affect this.

### DIFF
--- a/oauth-proxy/docker-compose.yml
+++ b/oauth-proxy/docker-compose.yml
@@ -26,4 +26,4 @@ services:
       - AWS_DEFAULT_REGION=us-west-2
     volumes:
       - .:/opt/app
-    command: "node index.js--config dev-config.json"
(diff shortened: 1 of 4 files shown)


oauth-proxy/docker-compose.yml
```diff
@@ -26,4 +26,4 @@ services:
       - AWS_DEFAULT_REGION=us-west-2
     volumes:
       - .:/opt/app
-    command: "--config dev-config.json"
+    command: "node build/app/js--config dev-config.json"
```
beaugrantham-va:
```
command: "node index.js --config dev-config.json"
```

dginther: noticed and fixed! :)

PR #105, opened 2020-05-28
Bug/saml proxy handle err

crolarlibertyva:
https://vajira.max.gov/browse/API-950

### DIFF
--- a/saml-proxy/docker-compose.yml
+++ b/saml-proxy/docker-compose.yml
@@ -6,5 +6,4 @@ services:
       dockerfile: saml-proxy/Dockerfile
     ports:
       - "7000:7000"
-    network_mode: "host"
     command: 'node build/app.js --config dev-config.json'
--- a/saml-proxy/views/handleFailure.hbs
+++ b/saml-proxy/views/handleFailure.hbs
@@ -2,7 +2,7 @@
 
 {{#>deptva-formation-error}}
   <h3 class="usa-alert-heading">
-    Some VA.gov tools aren't working right now
+    Some VA.gov service applications aren't working right now. Please note 
   </h3>
 
   <p>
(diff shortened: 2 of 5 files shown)


rtravitz:
If it's helpful, we have an example of an express default error handler we've been trying to improve in the developer-portal-backend. https://github.com/department-of-veterans-affairs/developer-portal-backend/blob/a161f029a96fe55122ed770630f90af3af5cd2d4/app.ts#L155

And the Sentry Express middleware: https://github.com/department-of-veterans-affairs/developer-portal-backend/blob/a161f029a96fe55122ed770630f90af3af5cd2d4/app.ts#L147

saml-proxy/docker-compose.yml
```diff
@@ -6,5 +6,4 @@ services:
       dockerfile: saml-proxy/Dockerfile
     ports:
       - "7000:7000"
-    network_mode: "host"
```
lukemarkwordtlibertyits, line 9: Why remove network_mode?

saml-proxy/views/handleFailure.hbs
```diff
@@ -2,7 +2,7 @@
 
 {{#>deptva-formation-error}}
   <h3 class="usa-alert-heading">
-    Some VA.gov tools aren't working right now
+    Some VA.gov service applications aren't working right now. Please note
```
beaugrantham-va, line 5: Not sure why this was changed... doesn't seem to add any value.

saml-proxy/docker-compose.yml
crolarlibertyva: When running in the host network mode the docker instance is not reachable via "localhost:7000" when running on docker for desktop in windows. Plus this made it consistent with oauth_proxy.

saml-proxy/views/handleFailure.hbs
crolarlibertyva: Will revert.

PR #109, opened 2020-06-05
Api 980 state parameter validation

lukemarkwordtlibertyits:
Added logic to the authorizationHandler and the redirectHandler to check for the state parameter. If the state parameter does not exist it will through an error message. Added unit tests to support this functionality. Refactored the oauthHandlers.test.js suite.


oauth-proxy/oauthHandlers/authorizeHandler.js
```diff
@@ -4,6 +4,15 @@ const { loginBegin } = require('../metrics');
 const authorizeHandler = async (config, redirect_uri, logger, issuer, dynamo, dynamoClient, oktaClient, req, res, next) => {
   loginBegin.inc();
   const { state, client_id, redirect_uri: client_redirect } = req.query;
+
+  if( state == null){
```
beaugrantham-va, line 8: Super minor comment, but let's make the formatting consistent with the current baseline. Same for the other file.
```
  if (state == null) {
```

PR #113, opened 2020-06-15
Api 1050 revoke endpoint

crolarlibertyva:


oauth-proxy/index.js
```diff
@@ -108,6 +114,26 @@ function buildApp(config, issuer, oktaClient, dynamo, dynamoClient, validateToke
...
   };
 
+  const proxyRevokeTokenRequestToOkta = (req, res, redirectUrl) => {
+    delete req.headers.host
+
+    axios({
+      method: "POST",
+      url: redirectUrl,
+      headers: req.headers,
+      data: querystring.stringify({
```
beaugrantham-va, line 124: I'm not sure I understand the need for this additional proxy handler... assuming the params come in the request body, why can't we use the proxy handler above?

crolarlibertyva: The outbound request with the other proxy call didn't format in a way that okta recognized a token parameter. Still not sure why, without looking at server-side (okta) logs for incoming request.

oauth-proxy/index.js
```diff
@@ -29,6 +31,7 @@ const openidMetadataWhitelist = [
   "token_endpoint",
   "userinfo_endpoint",
   "introspection_endpoint",
+  "revoke_endpoint",
```
beaugrantham-va, line 34: Let's make this consistent with Okta (`revocation_endpoint`), even though I can't find explicitly referenced it in the RFC.

We should also add `revocation_endpoint_auth_methods_supported` so the client knows how to to interact with it.

oauth-proxy/index.js
beaugrantham-va: This request (which I believe is valid) results in a 400 proxied back from Okta
```
curl -v -X POST \
  -u "$CLIENT_ID:$CLIENT_SECRET" \
  -d "token=$ACCESS_TOKEN" \
  "http://localhost:7100/oauth2/revoke"
```

oauth-proxy/index.js
```diff
@@ -130,6 +144,10 @@ function buildApp(config, issuer, oktaClient, dynamo, dynamoClient, validateToke
   }));
 
   router.use([appRoutes.token], bodyParser.urlencoded({ extended: true }));
+  router.use([appRoutes.revoke], bodyParser.urlencoded({ extended: true }));
```
beaugrantham-va, line 147: Let's choose to be explicit (`router.use`) or be global (`app.use`) and stick with that. Trying to figure out which applies in a mixed environment can be confusing.

oauth-proxy/index.js
```diff
@@ -86,17 +91,26 @@ function buildApp(config, issuer, oktaClient, dynamo, dynamoClient, validateToke
...
     targetResponse.status(response.status)
     response.data.pipe(targetResponse)
   };
 
-  const proxyRequestToOkta = (req, res, redirectUrl, requestMethod) => {
+  const proxyRequestToOkta = (req, res, redirectUrl, requestMethod, bodyencoder) => {
     delete req.headers.host
+    var payload = req.body;
+
+    // Only supporting x-www-form-urlencoded for now. json my be supported in the future if
```
beaugrantham-va, line 105: Is this comment still relevant assuming that the caller can now pass whatever bodyencoder they choose?

PR #116, opened 2020-06-29
API-1345 dependencies update for oauth2-proxy

crolarlibertyva:


oauth-proxy/package-lock.json
```diff
@@ -6538,28 +6186,67 @@
       "integrity": "sha1-HBH5IY8HYImkfdUS+TxmmaaoHVI="
     },
     "yargs": {
-      "version": "12.0.5",
-      "resolved": "https://registry.npmjs.org/yargs/-/yargs-12.0.5.tgz",
-      "integrity": "sha1-BfWZe2CWR7ZPZrgeO0sQo2jnrRM=",
+      "version": "15.3.1",
```
beaugrantham-va, line 6189: This major version change should also be made in `package.json`.

And being a major version, with potentially breaking changes, we need to do some really thorough testing.

oauth-proxy/package-lock.json
```diff
@@ -416,45 +436,35 @@
...
         "@types/istanbul-reports": "^1.1.1",
-        "@types/yargs": "^12.0.9"
+        "@types/yargs": "^13.0.0"
       }
     },
     "@okta/okta-sdk-nodejs": {
-      "version": "1.2.0",
-      "resolved": "https://registry.npmjs.org/@okta/okta-sdk-nodejs/-/okta-sdk-nodejs-1.2.0.tgz",
-      "integrity": "sha1-nDbr+H4XvaUOpLkvld/gTh1iA8s=",
+      "version": "3.3.1",
```
beaugrantham-va, line 456: This major version change should also be made in `package.json`.

And being a major version, with potentially breaking changes, we need to do some really thorough testing.

PR #118, opened 2020-07-07
Api 1304 add revoke okta grants enpoint

lukemarkwordtlibertyits:
[API1304_TestCases.docx](https://github.com/department-of-veterans-affairs/vets-saml-proxy/files/4887372/API1304_TestCases.docx)
Created Delete /grants endpoint which revokes a user's okta grants on a particular client.


oauth-proxy/apiClients/oktaApiClient.js
```diff
@@ -0,0 +1,23 @@
+const axios = require('axios');
+const uriTemplates = require('uri-templates');
+
+const deleteUserGrantOnClient = async (config, userId, ssws, clientId) => {
```
beaugrantham-va, line 4: `config` is already a parameter here, so I don't see a need to also have `ssws` separately.

oauth-proxy/apiClients/oktaApiClient.js
```diff
@@ -0,0 +1,23 @@
...
+  let error;
+  let response;
+  const template = uriTemplates(config.okta_url+"/api/v1/users/{userid}/clients/{clientid}/grants")
+  await axios({
+      method: "DELETE",
+      url: template.fill({userid: userId, clientid: clientId}),
```
beaugrantham-va, line 10: The story references users by email address, but `userId` looks to be the Okta assigned identifier.

In terms of usability, I'm not sure a client will immediately know the Okta assigned identifier. They would have to pull it ahead of time from the `/userinfo` endpoint using a valid token.

Alternatively, we'd have to pull it via the API if given an email address.

I'm down to discuss the best plan and update the story if needed.

oauth-proxy/index.js
```diff
@@ -195,6 +196,10 @@ function buildApp(config, issuer, oktaClient, dynamo, dynamoClient, validateToke
       .catch(next)
   });
 
+  router.delete(appRoutes.grants, async (req, res, next) => {
+    oauthHandlers.revokeUserGrantHandler(config, req, res, next).catch(next)
```
beaugrantham-va, line 200: Should there be an `await` here since `revokeUserGrantHandler` is `async`?

beaugrantham-va:
The sample config file (`dev-config.base.json`) and `cli.js` should be updated with the new config option.

oauth-proxy/apiClients/oktaApiClient.js
lukemarkwordtlibertyits: The user_id is now fetched in the `revokeUserGrantHandler` and is passed into this method. End Users using the Delete /grants endpoint will only need to pass in an email address.

oauth-proxy/apiClients/oktaApiClient.js
lukemarkwordtlibertyits: This has been corrected.

oauth-proxy/index.js
lukemarkwordtlibertyits: This has been resolved.

oauth-proxy/oauthHandlers/revokeUserGrantHandler.js
```diff
@@ -0,0 +1,129 @@
...
+    if(clientError){
+        throw {"status": clientError.response.status, "errorMessage": clientError.response.data.error_description}
+    }
+}
+
+const checkIfParamsExist = (clientId, email) => {
+    let errorMessage = "";
+
+    if(!clientId || clientId == ""){
+        errorMessage += "Client Id is a required parameter. "
```
beaugrantham-va, line 112: These error messages should reference the actual parameter name (e.g. `client_id` vs `Client Id`).

oauth-proxy/oauthHandlers/revokeUserGrantHandler.js
```diff
@@ -0,0 +1,129 @@
...
+
+    try {
+        userIds = await getUserIds(config, email);
+    }catch(error){
+        setErrorResponse(res, error.status, error.errorMessage);
+        return next();
+    }
+
+    if(userIds.length < 1){
+        setErrorResponse(res, 404, "No Okta userIds associated with the email: "+email);
```
beaugrantham-va, line 25: Let's drop _Okta_ from the error response, since a client may or may not be aware of it.

oauth-proxy/apiClients/oktaApiClient.js
```diff
@@ -0,0 +1,67 @@
...
+  if(response == null){
+    throw error;
+  }
+  
+  return response;
+}
+
+const getUserInfo = async (config, email) => {
+  //Assume that the email is passed into this method unencoded.
+  let uri = URI(config.okta_url+"/api/v1/users");
```
beaugrantham-va, line 26: I think this API is too permissive for looking up users.

> Use the `q` parameter for a simple lookup of users by name, for example when creating a people picker. The value of `q` is matched against `firstName`, `lastName`, or `email`.
> Performs a startsWith match but this is an implementation detail and may change without notice. You don't need to specify `firstName`, `lastName`, or `email`.

So if I pass `a` as the `email`, I get many many records.

It's also not safe to assume the value is encoded at this point. It would have been decoded by the application and will need to be re-encoded to be safe.

lukemarkwordtlibertyits: Thats a good point. Do you think I should include a check that the parameter is a valid email?

lukemarkwordtlibertyits: I'll explore some stuff first.

beaugrantham-va: I see two options that might be better.
1. [Get User with Login](https://developer.okta.com/docs/reference/api/users/#get-user-with-login) if we are storing the email address as the login
2. [List Users with a Filter](https://developer.okta.com/docs/reference/api/users/#list-users-with-a-filter) that can take a `profile.email` filter.

lukemarkwordtlibertyits: I am exploring the profile.email option right now. It seems like it will work.

lukemarkwordtlibertyits: I have implemented the filter

oauth-proxy/oauthHandlers/revokeUserGrantHandler.js
lukemarkwordtlibertyits: Done.

oauth-proxy/oauthHandlers/revokeUserGrantHandler.js
lukemarkwordtlibertyits: Done.

oauth-proxy/apiClients/oktaApiClient.js
```diff
@@ -0,0 +1,68 @@
...
+  
+  if(response == null){
+    throw error;
+  }
+  
+  return response;
+}
+
+const getUserInfo = async (config, email) => {
+  //Assume that the email is passed into this method unencoded.
```
beaugrantham-va, line 25: This comment is not quite accurate.

The value of `email` may or may not have been encoded when it was sent by the client, but at this point it will have been decoded.

It gets encoded below (`uri.search`) before going out again.
> Note that names and values passed in an object are encoded automatically.
> https://medialize.github.io/URI.js/docs.html#accessors-search

So I think it's safe to remove this comment.

lukemarkwordtlibertyits: done.

PR #125, opened 2020-08-04
Development error handler now returns a generic 400 response

lukemarkwordtlibertyits:


saml-proxy/src/routes/index.js
```diff
@@ -161,13 +161,11 @@ export default function configureExpress(app, argv, idpOptions, spOptions, vetsA
     }));
   }
 
-  // development error handler
```
lukemarkwordtlibertyits, line 164: There is no switch for this feature. It is available in all environments at the moment. Although not advertised, there is nothing stopping a 3rd party from creating a development client, and Developer will have access to the `saml-proxy` logs. I decided it was probably just best to remove this feature as a whole.

beaugrantham-va: Suppressing stack traces and detailed error reports is a good idea. I'm onboard with that.

I'm not so sure that claiming _all_ errors are client induced (400) and not providing _any_ detail is a good idea. This will hide actual bugs (5xx) from surfacing and will throw off our dashboards and alerts.

If the error is 4xx it should include some indication of what the problem is so the client can fix it and try again... e.g. bad XML. I think `err.message` should generally capture this?

If the error is 5xx a generic error message is good.

lukemarkwordtlibertyits: Agreed `err.message` should be enough. I'll get the commit for that out shortly.

PR #129, opened 2020-08-14
Api 1949 add linting to oauth proxy

lukemarkwordtlibertyits:
Added automated formatting and linting to oauth-proxy. Applied suggested changes.

[Execution and Regression Tests](https://github.com/department-of-veterans-affairs/vets-saml-proxy/files/5075196/API1949_Execution.docx)

### DIFF
--- a/oauth-proxy/Dockerfile
+++ b/oauth-proxy/Dockerfile
@@ -19,6 +19,8 @@ EXPOSE 7100 7100
 HEALTHCHECK --interval=1m --timeout=4s --start-period=30s \
   CMD node bin/healthcheck.js
 
+CMD npm run-script lint
+
 USER node
 ENTRYPOINT ["/usr/local/bin/tini", "--"]
 CMD ["node", "index.js"]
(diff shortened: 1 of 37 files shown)


oauth-proxy/tests/e2e.test.js
```diff
@@ -1,562 +1,649 @@
...
+					},
+					auth: { username: "clientId123", password: "secretXyz" },
+				}
+			)
+			.then((resp) => {
+				expect(resp.status).toEqual(200);
+				expect(resp.data.username).toEqual("john.doe@example.com");
+			})
+			.catch(() => {
+				expect(true).toEqual(false);
```
lukemarkwordtlibertyits, line 422: I added a couple of these statements to the introspect tests, before there were no assertions in the catch statement. Since the assertions are in the then statement they would get thrown to the catch and pass.

oauth-proxy/Dockerfile
```diff
@@ -19,6 +19,8 @@ EXPOSE 7100 7100
 HEALTHCHECK --interval=1m --timeout=4s --start-period=30s \
   CMD node bin/healthcheck.js
 
+run npm run-script lint
```
crolarlibertyva, line 22: This belong here and in the Jenkinsfile too? Though I do like seeing the results in the docker build.

lukemarkwordtlibertyits: Beau and I decided to take it out.

PR #131, opened 2020-08-14
API-1741 - Log audience

lukemarkwordtlibertyits:
Audit audience parameter passed to `/authorization`.

### DIFF
--- a/oauth-proxy/oauthHandlers/authorizeHandler.js
+++ b/oauth-proxy/oauthHandlers/authorizeHandler.js
@@ -56,7 +56,7 @@ const checkParameters = async (state, aud, config, issuer, logger) => {
   if(aud) {
     let authorizationServerId = new URL(issuer.metadata.issuer).pathname.split('/').pop();
     let serverAudiences;
-    
+
     await getAuthorizationServerInfo(config, authorizationServerId)
     .then(res => {
       serverAudiences = res.audiences;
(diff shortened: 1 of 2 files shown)


oauth-proxy/oauthHandlers/authorizeHandler.js
```diff
@@ -62,11 +62,17 @@ const checkParameters = async (state, aud, config, issuer) => {
       serverAudiences = res.audiences;
     })
     .catch(() => {
-      throw {status: 500, error: "internal_error"};
```
beaugrantham-va, line 65: I think this should continue to be an error condition.

oauth-proxy/oauthHandlers/authorizeHandler.js
```diff
@@ -62,11 +62,17 @@ const checkParameters = async (state, aud, config, issuer) => {
...
     });
 
+    if(!serverAudiences){
+      return;
+    }
+
     if(!serverAudiences.includes(aud)){
-      throw {status: 400, error: "invalid_request", error_description: "Invalid aud parameter"};
+      /* throw {status: 400, error: "invalid_request", error_description: "Invalid aud parameter"};*/
+      logger.info("Client included invalid audience parameter.")
```
beaugrantham-va, line 75: This should be a `warn` and it should include the actual vs expected values in the log.

crolarlibertyva:
Warning works as intended from my local testing.

PR #140, opened 2020-09-03
API-1853: Support oauth-proxy ECS in Staging 

mramirez-va:
We've added support for generating the `config.json` file required by the `oauth-proxy` app. The dockerfile for the deployment into Fargate is being updated to exercise this change.

https://vajira.max.gov/browse/API-1853

https://github.com/department-of-veterans-affairs/health-apis-docker-octopus/pull/34


duganth-va (approved):
Looks good

PR #141, opened 2020-09-04
API-1853: Support oauth-proxy ECS in Staging (continued)

mramirez-va:
We've added support for generating the config.json file required by the oauth-proxy app. The dockerfile for the deployment into Fargate is being updated to exercise this change. 

- Override the command to point to the newly generated `config.json` file

https://vajira.max.gov/browse/API-1853

department-of-veterans-affairs/health-apis-docker-octopus#34


duganth-va (approved):
lgtm

PR #145, opened 2020-09-17
API-2338-update-formation-dependency Updated formation from 3.0.0 to …

bdgrider:
…6.9.5

Updated the dependency @department-of-veterans-affairs/formation from 3.0.0 to 6.9.5. 

https://vajira.max.gov/browse/API-2338

### DIFF
--- a/saml-proxy/package-lock.json
+++ b/saml-proxy/package-lock.json
@@ -407,19 +407,20 @@
       }
     },
     "@department-of-veterans-affairs/formation": {
-      "version": "3.0.0",
-      "resolved": "https://registry.npmjs.org/@department-of-veterans-affairs/formation/-/formation-3.0.0.tgz",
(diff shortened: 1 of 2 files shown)


beaugrantham-va:
Were there any UI/styling changes as a result of this since the dependency affects CSS?

crolarlibertyva:
> Were there any UI/styling changes as a result of this since the dependency affects CSS?

Change seems good so far. When i did a run, I didn't notice any changes in the rendering of the saml-proxy page. I'll try again and look more closely though.

crolarlibertyva (approved):
Looks good to me. Authenticates correctly in my local setup.

bdgrider:
Would that just be the authentication page accessible via localhost:7000? @beaugrantham-va It appears to be unchanged.

bdgrider:
I just verified nothing on the auth page appears different or broken.

bdgrider:
The CSS appears unchanged but the auth flow is not working as it should. I am getting this matching error after entering what I believe to be correct log in information @beaugrantham-va 

image

bdgrider:
After regression testing the saml-proxy after updating this dependency, there appeared to be no style changes that occurred. All functionality was exercised and is in working order.

lukemarkwordtlibertyits (approved):
My tests passed. Looks good to me.

PR #146, opened 2020-09-18
Api 2333 update dependencies

bdgrider:
After regression testing this ticket, it looks like doing the recommended npm installation (npm install uswds@2.8.1) to fix the first yargs-parser vulnerability actually breaks functionality in the saml-proxy service. After rolling back to the original version of uswds (1.6.8) the error no longer occurs. 

I took care of the other yargs-parser vulnerability by updating ts-jest to version 26.3.0. Submitting a PR as is for now.

### DIFF
--- a/saml-proxy/package.json
+++ b/saml-proxy/package.json
@@ -93,14 +93,14 @@
     "request": "^2.88.0",
     "request-promise-native": "^1.0.5",
     "samlp": "git+https://github.com/edpaget/node-samlp.git",
-    "uswds": "^2.8.1",
+    "uswds": "^1.6.8",
     "uuid": "^3.3.2",
     "winston": "^3.3.3",
     "xml-formatter": "^1.0.1",
     "xml2js": "^0.4.19",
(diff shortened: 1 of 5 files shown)


saml-proxy/package.json
```diff
@@ -99,7 +99,8 @@
     "xml-formatter": "^1.0.1",
     "xml2js": "^0.4.19",
     "xmldom": "^0.1.27",
-    "yargs": "^15.3.1"
+    "yargs": "^16.0.3",
```
crolarlibertyva, line 102: Just to make sure, the ticket is for both saml-proxy and the proxy-monitor?

crolarlibertyva: So far good with the saml-proxy.

crolarlibertyva: Good with proxy-monitor.

crolarlibertyva (approved):

lukemarkwordtlibertyits (approved):

PR #147, opened 2020-09-21
Api 2418 oauth proxy update node fetch

lukemarkwordtlibertyits:
[API2418_Tests.docx](https://github.com/department-of-veterans-affairs/vets-saml-proxy/files/5258537/API2418_Tests.docx)

### DIFF
--- a/oauth-proxy/package-lock.json
+++ b/oauth-proxy/package-lock.json
@@ -1694,24 +1694,6 @@
       "resolved": "https://registry.npmjs.org/encodeurl/-/encodeurl-1.0.2.tgz",
       "integrity": "sha1-rT/0yG7C0CkyL1oCw6mmBslbP1k="
     },
-    "encoding": {
-      "version": "0.1.13",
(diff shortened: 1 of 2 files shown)


crolarlibertyva (commented):
@lukemarkwordtlibertyits : Do we know if there's an updated docker image to reference in the Dockerfile?

crolarlibertyva:
> @lukemarkwordtlibertyits : Do we know if there's an updated docker image to reference in the Dockerfile?

Nevermind. It's already set to the latest tag.

crolarlibertyva (commented):
Runs good for me.

dbrown-va (approved):
Confirmed both the initial OAuth flow and ability to obtain a refresh token, so functionally I think this is good to go.

Avoiding npm force resolutions would be preferable from a maintenance perspective, so let's hold on the merge until we sync up.

beaugrantham-va:
Closing pending parent dependency updates and/or action with developer-portal-backend.

PR #151, opened 2020-09-28
API-2512 - Update node-forge dependency

bdgrider:
Committing package.json and package-lock.json after verifying authentication flow is fully functional.

https://vajira.max.gov/browse/API-2512

### DIFF
--- a/saml-proxy/package.json
+++ b/saml-proxy/package.json
@@ -84,6 +84,7 @@
     "lodash.omit": "^4.5.0",
     "lodash.template": "^4.5.0",
     "morgan": "~1.9.0",
+    "node-forge": "^0.10.0",
     "node-sass": "^4.14.1",
     "node-sass-middleware": "^0.11.0",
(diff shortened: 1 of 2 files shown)


beaugrantham-va (changes requested):
I'm not sure this fixes the problem.

It looks like there are still references to 0.7.6 child dependencies. I also don't think this should be a top level dependency unless we actually need it there.

saml-proxy/package.json
```diff
@@ -84,6 +84,7 @@
     "lodash.omit": "^4.5.0",
     "lodash.template": "^4.5.0",
     "morgan": "~1.9.0",
+    "node-forge": "^0.10.0",
```
crolarlibertyva, line 87: If this is a transitive dependency, we should make the change in package-lock.json.

beaugrantham-va:
Closing pending parent dependency updates and/or action with developer-portal-backend.

PR #153, opened 2020-09-30
Api 2513 node forge update

bdgrider:
Got rid of the node-forge nested dependency by updating openid-client to 3.15.10 (4.1.1 caused an error to occur during docker-compose usage, so I used the latest version of the previous major version which got the baseline back to a functional state).


crolarlibertyva (approved):
Successfully run on my local setup. Confirmed the removal of the high priority vulnerability.

beaugrantham-va (approved):

PR #159, opened 2020-10-13
API-2249 - Reject missing redirect

bdgrider:
Logged and threw error message for the case of when someone is trying to authenticate without a relay state.

https://vajira.max.gov/browse/API-2249

### DIFF
--- a/saml-proxy/src/routes/handlers.js
+++ b/saml-proxy/src/routes/handlers.js
@@ -112,6 +112,11 @@ export const parseSamlRequest = function (req, res, next) {
         };
       }
       req.session.authnRequest = req.authnRequest;
+    } else {
+      logger.error("Empty request data. Invalid request.");
+      throw {
+        message: "Error: Empty request data. Invalid request.",
+      };
     }
     next();
   });
--- a/saml-proxy/src/routes/handlers.js
+++ b/saml-proxy/src/routes/handlers.js
@@ -102,6 +102,15 @@ export const parseSamlRequest = function (req, res, next) {
         acsUrl: data.assertionConsumerServiceURL,
         forceAuthn: data.forceAuthn === "true",
       };
+      if (
+        req.authnRequest.relayState == null ||
+        req.authnRequest.relayState == ""
+      ) {
+        logger.error("Empty relay state. Invalid request.");
+        throw {
+          message: "Error: Empty relay state. Invalid request.",
+        };
+      }
       req.session.authnRequest = req.authnRequest;
     }
     next();
--- a/saml-proxy/src/routes/handlers.js
+++ b/saml-proxy/src/routes/handlers.js
@@ -112,11 +112,6 @@ export const parseSamlRequest = function (req, res, next) {
         forceAuthn: data.forceAuthn === "true",
       };
       req.session.authnRequest = req.authnRequest;
-    } else if (!data && !req.authnRequest) {
-      logger.error("Empty request data. Invalid request.");
-      throw {
-        message: "Error: Empty request data. Invalid request.",
-      };
     }
     next();
   });
--- a/saml-proxy/src/routes/handlers.js
+++ b/saml-proxy/src/routes/handlers.js
@@ -32,6 +32,15 @@ export const samlLogin = function (template) {
       ? req.authnRequest
       : req.session.authnRequest;
     req.authnRequest = authnRequest;
+    if (
+      req.authnRequest.relayState == null ||
+      req.authnRequest.relayState == ""
+    ) {
+      logger.error("Empty relay state. Invalid request.");
+      throw {
+        message: "Error: Empty relay state. Invalid request.",
+      };
+    }
     const samlp = new _samlp(
       req.sp.options.getResponseParams(),
       new SAML.SAML(req.sp.options.getResponseParams())
@@ -102,17 +111,8 @@ export const parseSamlRequest = function (req, res, next) {
         acsUrl: data.assertionConsumerServiceURL,
         forceAuthn: data.forceAuthn === "true",
       };
-      if (
-        req.authnRequest.relayState == null ||
-        req.authnRequest.relayState == ""
-      ) {
-        logger.error("Empty relay state. Invalid request.");
-        throw {
-          message: "Error: Empty relay state. Invalid request.",
-        };
-      }
       req.session.authnRequest = req.authnRequest;
-    } else {
+    } else if (!data && !req.authnRequest) {
       logger.error("Empty request data. Invalid request.");
       throw {
         message: "Error: Empty request data. Invalid request.",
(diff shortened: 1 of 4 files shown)


beaugrantham-va (changes requested):
I'm not sure this is actually working.

I can visit /samlproxy/idp/saml/sso and don't get an error, even though the logs clearly show that relay state / body is undefined.

```
saml-proxy_1  | {
saml-proxy_1  |   "service": "saml-proxy",
saml-proxy_1  |   "request_id": "e92df540-0d9a-11eb-8be2-7935f568e531",
saml-proxy_1  |   "time": "2020-10-13T21:27:35.317Z",
saml-proxy_1  |   "step": "from Okta",
saml-proxy_1  |   "session": "ee8742f2-35bb-4c26-9046-2f2a53f5f9e5",
saml-proxy_1  |   "level": "info",     
saml-proxy_1  |   "message": "Relay state from Okta - body: undefined query: undefined"
saml-proxy_1  | }                                                                                                          
saml-proxy_1  | {
saml-proxy_1  |   "service": "saml-proxy",
saml-proxy_1  |   "request_id": "e92df540-0d9a-11eb-8be2-7935f568e531",
saml-proxy_1  |   "action": "parseSamlRequest",      
saml-proxy_1  |   "result": "success",
saml-proxy_1  |   "session": "ee8742f2-35bb-4c26-9046-2f2a53f5f9e5",
saml-proxy_1  |   "level": "info",     
saml-proxy_1  |   "message": "User arrived from Okta. Rendering IDP login template.",
saml-proxy_1  |   "time": "2020-10-13T21:27:35.329Z"                                                                       
saml-proxy_1  | }
saml-proxy_1  | {
saml-proxy_1  |   "remote-address": "::ffff:172.20.0.1",
saml-proxy_1  |   "time": "2020-10-13T21:27:35.334Z",
saml-proxy_1  |   "method": "GET",
saml-proxy_1  |   "url": "/samlproxy/idp/saml/sso",                                                                        
saml-proxy_1  |   "status-code": "200",
saml-proxy_1  |   "content-length": "7310",                                                                                
saml-proxy_1  |   "request_id": "e92df540-0d9a-11eb-8be2-7935f568e531"
saml-proxy_1  | }
```

I'd also like to see a unit test for this.

lukemarkwordtlibertyits:
This changes breaks the verify endpoint

saml-proxy/src/routes/handlers.js
```diff
@@ -102,7 +102,21 @@ export const parseSamlRequest = function (req, res, next) {
         acsUrl: data.assertionConsumerServiceURL,
         forceAuthn: data.forceAuthn === "true",
       };
+      if (
+        req.authnRequest.relayState == null ||
```
lukemarkwordtlibertyits, line 106: This check should go in the `samlLogin` function

saml-proxy/src/routes/handlers.js
```diff
@@ -102,7 +102,21 @@ export const parseSamlRequest = function (req, res, next) {
...
+        req.authnRequest.relayState == ""
+      ) {
+        logger.error("Empty relay state. Invalid request.");
+        throw {
+          message: "Error: Empty relay state. Invalid request.",
+        };
+      }
       req.session.authnRequest = req.authnRequest;
+    } else {
```
lukemarkwordtlibertyits, line 115: This else statement along with the relay state check will break the `verify` endpoint.

saml-proxy/src/routes/handlers.js
lukemarkwordtlibertyits: right after [this](https://github.com/department-of-veterans-affairs/vets-saml-proxy/pull/159/files#diff-31b89a2353a0eafd6b5da7f00254c006264532e90dee2065bf45018dcda549deR34) statement.

lukemarkwordtlibertyits (changes requested):

saml-proxy/src/routes/handlers.js
lukemarkwordtlibertyits: The reason is that this method creates an authnRequest object that is used to make a saml request to our IDPs. The object is stored in session, so it is possible to hit the `idp/saml/sso` endpoint without an authnRequest or relayState query parameter, so long as these parameters are stored in the request session.

crolarlibertyva:
Linting items need to be addressed.

crolarlibertyva (changes requested):
Need to address lint violations.

saml-proxy/src/routes/handlers.js
```diff
@@ -103,6 +112,11 @@ export const parseSamlRequest = function (req, res, next) {
         forceAuthn: data.forceAuthn === "true",
       };
       req.session.authnRequest = req.authnRequest;
+    } else if (!data && !req.authnRequest) {
```
lukemarkwordtlibertyits, line 115: The parseSamlRequest is called before the samlLogin method so this check will break the /verify endpoint. There are use cases for parseSamlRequest to return null. I believe this check should be removed entirely, until we decide whether we would like to refactor this module. For now, a single check in the samlLogin method should be enough.
[API2249_Tests.docx](https://github.com/department-of-veterans-affairs/vets-saml-proxy/files/5394690/API2249_Tests.docx)

saml-proxy/src/routes/handlers.js
```diff
@@ -32,6 +32,15 @@ export const samlLogin = function (template) {
       ? req.authnRequest
       : req.session.authnRequest;
     req.authnRequest = authnRequest;
+    if (
```
lukemarkwordtlibertyits, line 35: This the empty data check in the parseSamlRequest method should be taken out this check should also include a null check on the authnRequest object.
ex. `req.authnRequest?.relayState == null`

lukemarkwordtlibertyits (changes requested):

lukemarkwordtlibertyits (approved):

lukemarkwordtlibertyits:
Looks Good!
[API2249_Tests__V2.docx](https://github.com/department-of-veterans-affairs/vets-saml-proxy/files/5403099/API2249_Tests__V2.docx)

saml-proxy/src/routes/handlers.js
```diff
@@ -32,6 +32,16 @@ export const samlLogin = function (template) {
...
+      req.authnRequest?.relayState == null ||
```
beaugrantham-va, line 36: I question whether this should be checking for `null`, or whether it should be checking for `=== undefined`.

I don't see anywhere in the code where we explicitly set this to `null`.

lukemarkwordtlibertyits: In javascript null defines an object whereas undefined defines a type, so `null == undefined //true` but `null === undefined \\false`.
So this check should cover both nulls and undefined.

beaugrantham-va (approved):

PR #160, opened 2020-10-19
Resolved outdated node-fetch dependency

lukemarkwordtlibertyits:
The okta-sdk had an outdated isomorphic-fetch dependency. This dependency had an outdated node-fetch dependency that caused dependabot to report a minor security vunerability. Okta-SDK's latest patch updated the isomorphic-fetch dependency version to the latest version, which resolved its node-fetch issue.
[Regression Tests](https://github.com/department-of-veterans-affairs/vets-saml-proxy/files/5403530/API2418_Tests__RegressionTests-V2.docx)


crolarlibertyva (changes requested):
I updated this dependency during API-2399.  This could be the reason for the conflict.

oauth-proxy/package.json
```diff
@@ -22,7 +22,7 @@
   },
   "homepage": "https://github.com/department-of-veterans-affairs/oauth-proxy#readme",
   "dependencies": {
-    "@okta/okta-sdk-nodejs": "^3.3.1",
+    "@okta/okta-sdk-nodejs": "^4.2.0",
```
crolarlibertyva, line 25: I updated this dependency during API-2399.  This could be the reason for the conflict.

lukemarkwordtlibertyits: yes, just resolved.

lukemarkwordtlibertyits:
The rebase merged in a weird way. I feel more comfortable creating a new branch for this PR.

PR #162, opened 2020-10-20
Api 000 refactor unit tests

lukemarkwordtlibertyits:
Before taking on API-3042 (refactoring the token handler). I thought it would be helpful to break the oauth-proxy unit tests into separate module, based on the handler they are testing.  This will make finding and reading unit tests easier in the future. Also putting all the test helper methods will hopefully reduce repeated code in the future.


crolarlibertyva (approved):
I like this a lot better.

beaugrantham-va (approved):

PR #163, opened 2020-10-22
API-2890

bdgrider:
Moving oauth proxy jest config to package.json. Verified coverage threshold behavior works with test:cli script.
[API-2890_test.txt](https://github.com/department-of-veterans-affairs/vets-saml-proxy/files/5424645/API-2890_test.txt)


oauth-proxy/package.json
```diff
@@ -5,12 +5,43 @@
...
     "help": "node index.js --help"
   },
+  "jest": {
+    "clearMocks": true,
+    "collectCoverage": true,
+    "testEnvironment": "node",
+    "testMatch": ["**/*.test.js"],
+    "coverageThreshold": {
+      "global": {
+        "lines": 74
```
crolarlibertyva, line 22: So this is number of lines or percentage?

saml-proxy/package.json
```diff
@@ -32,6 +32,12 @@
     "dev": "./node_modules/nodemon/bin/nodemon.js"
   },
   "jest": {
+    "collectCoverage": true,
+    "coverageThreshold": {
+      "global": {
+        "lines": 77
```
crolarlibertyva, line 38: So this is number of lines or percentage? Same as above.

oauth-proxy/package.json
```diff
@@ -36,6 +67,8 @@
     "node-cache": "^5.1.2",
     "openid-client": "^3.15.10",
     "prom-client": "^11.5.2",
+    "ts-jest": "^26.4.1",
```
beaugrantham-va, line 70: Is all of this `ts` stuff supposed to be on the saml-proxy project? There is no TypeScript in the oauth-proxy and I see a bunch of warnings about it when running `npm test`.

oauth-proxy/package.json
beaugrantham-va: > Thresholds specified as a positive number are taken to be the minimum percentage required.

https://jestjs.io/docs/en/configuration#coveragethreshold-object

saml-proxy/package.json
beaugrantham-va: > Thresholds specified as a positive number are taken to be the minimum percentage required.

https://jestjs.io/docs/en/configuration#coveragethreshold-object

oauth-proxy/package.json
bdgrider: Yes, you are right. I just removed the typescript node modules and removed all references to it in the package.json. Accidentally copied these over when syncing up the config so that they will look mostly the same.

oauth-proxy/package.json
```diff
@@ -5,12 +5,40 @@
...
+      "/node_modules/",
+      "/build/"
+    ],
+    "moduleFileExtensions": [
+      "js",
+      "json",
+      "node"
+    ],
+    "globals": {
+      "ts-jest": {
```
beaugrantham-va, line 37: There are still some ts references.

oauth-proxy/package.json
```diff
@@ -36,6 +64,7 @@
     "node-cache": "^5.1.2",
     "openid-client": "^3.15.10",
     "prom-client": "^11.5.2",
+    "typescript": "^3.9.7",
```
beaugrantham-va, line 67: And here.

beaugrantham-va (changes requested):

beaugrantham-va (approved):

crolarlibertyva (approved):

PR #164, opened 2020-10-22
Api 3042 refactor token handler

lukemarkwordtlibertyits:
Refactored Token Handler.
[API3042_Tests__RegressionTests.docx](https://github.com/department-of-veterans-affairs/vets-saml-proxy/files/5426463/API3042_Tests__RegressionTests.docx)


oauth-proxy/oauthHandlers/tokenHandlerStrategyClasses/unsupportedGrantStrategy.js
```diff
@@ -0,0 +1,14 @@
+class UnsupportedGrantStrategy {
+  constructor() {}
+
+  //will throw error if cannot retrieve refresh token
+  async getToken(client, redirect_uri) { // eslint-disable-line
+    throw {
```
beaugrantham-va, line 6: This previously returned a different error.

```
res.status(400).json({
  error: "unsupported_grant_type",
  error_description:
    "Only authorization and refresh_token grant types are supported",
});
```

oauth-proxy/oauthHandlers/tokenHandlerStrategyClasses/unsupportedGrantStrategy.js
```diff
@@ -0,0 +1,14 @@
+class UnsupportedGrantStrategy {
+  constructor() {}
+
+  //will throw error if cannot retrieve refresh token
```
beaugrantham-va, line 4: This comment is not applicable here.

oauth-proxy/oauthHandlers/tokenHandlerStrategyClasses/tokenHandlerClient.js
```diff
@@ -0,0 +1,114 @@
...
+
+    let tokens = await this.tokenHandlerStrategy.getToken(
+      client,
+      this.redirect_uri
+    );
+
+    let document = await this.tokenHandlerStrategy.pullDocumentFromDynamo();
+    let state;
+    if (document && tokens) {
+      this.tokenHandlerStrategy.saveDocumentToDynamo(document, tokens);
```
beaugrantham-va, line 45: This is missing `await`.

oauth-proxy/oauthHandlers/tokenHandlerStrategyClasses/unsupportedGrantStrategy.js
lukemarkwordtlibertyits: Good catch

dbrown-va:
Since this is a public repo, it could be beneficial to add a bit more info to the PR description for the rational behind the PR, especially for those who may not have JIRA access to look up the issue.

Something like..

> This PR breaks up the single token handler into smaller units with dedicated function.  List function(s) and their purposes, etc.

oauth-proxy/oauthHandlers/tokenHandler.js
```diff
@@ -21,150 +23,69 @@ const tokenHandler = async (
...
-      );
-      const statusCode = statusCodeFromError(error);
-      res.status(statusCode).json({
-        error: error.error,
-        error_description: error.error_description,
+  let tokenResponse;
+  try {
+    tokenResponse = await tokenHandlerClient.handleToken();
+  } catch (err) {
+    if (err.name == "TokenHandlerError" && !err.sentry) {
```
lukemarkwordtlibertyits, line 52: Now we must specify that we do not want the error sent to sentry. We do this by throwing the new `TokenHandlerError` class which has a sentry field, that defaults to false. If we want to handle an error but still send it to sentry we can change the sentry field to true.
 Any error that is not a `TokenHandlerError` will be assumed to be an unhandled error, therefore we would send that error to sentry.

lukemarkwordtlibertyits: Decided against this. Added more handling in the TokenHandlerClient.

oauth-proxy/oauthHandlers/tokenHandlerStrategyClasses/tokenHandlerClient.js
```diff
@@ -0,0 +1,137 @@
...
+    let document = await this.tokenHandlerStrategy.pullDocumentFromDynamo();
+    let state;
+    if (document && tokens) {
+      await this.tokenHandlerStrategy.saveDocumentToDynamo(document, tokens);
+      state = document.state.S;
+    }
+
+    //Creates a Token Response
+    const tokenResponseBase = translateTokenSet(tokens);
+    var decoded = jwtDecode(tokens.access_token);
```
beaugrantham-va, line 74: My IDE is recommending the following change.

> 'var' used instead of 'let' or 'const'

oauth-proxy/oauthHandlers/tokenHandlerStrategyClasses/tokenHandlerClient.js
```diff
@@ -0,0 +1,137 @@
...
+    try {
+      const validation_result = await this.validateToken(
+        tokens.access_token,
+        decoded.aud
+      );
+      patient = validation_result.va_identifiers.icn;
+    } catch (error) {
+      rethrowIfRuntimeError(error);
+      let returnError = {
+        error: "invalid_grant",
```
beaugrantham-va, line 126: Just want to confirm... this previously threw a 400 error and now it bubbles up to the default error handler and returns a 500?

I do believe 500 is the correct response here.

oauth-proxy/oauthHandlers/tokenHandlerStrategyClasses/refreshTokenStrategy.js
```diff
@@ -0,0 +1,70 @@
...
+const { rethrowIfRuntimeError, statusCodeFromError } = require("../../utils");
+const { oktaTokenRefreshGauge, stopTimer } = require("../../metrics");
+
+class RefreshTokenStrategy {
+  constructor(req, logger, dynamo, dynamoClient) {
+    this.req = req;
+    this.logger = logger;
+    this.dynamo = dynamo;
+    this.dynamoClient = dynamoClient;
+    this.oktaTokenRefreshStart;
```
beaugrantham-va, line 11: This should be local to `getToken`

oauth-proxy/oauthHandlers/tokenHandlerStrategyClasses/refreshTokenStrategy.js
```diff
@@ -0,0 +1,70 @@
+const process = require("process");
```
beaugrantham-va, line 1: My IDE is reporting:

> Module is not listed in package.json dependencies 

I think this must have been a problem before this PR but let's go ahead and fix it while we're in here.

oauth-proxy/oauthHandlers/tokenHandlerStrategyClasses/refreshTokenStrategy.js
```diff
@@ -0,0 +1,70 @@
...
+  constructor(req, logger, dynamo, dynamoClient) {
+    this.req = req;
+    this.logger = logger;
+    this.dynamo = dynamo;
+    this.dynamoClient = dynamoClient;
+    this.oktaTokenRefreshStart;
+  }
+
+  //will throw error if cannot retrieve refresh token
+  async getToken(client, redirect_uri) { // eslint-disable-line
```
beaugrantham-va, line 15: I'm a little concerned about unused params, but I don't know how it's meant to be solved with this design pattern.

oauth-proxy/oauthHandlers/tokenHandlerStrategyClasses/refreshTokenStrategy.js
```diff
@@ -0,0 +1,70 @@
...
+        "refresh_token",
+        this.req.body.refresh_token
+      );
+    } catch (error) {
+      this.logger.error("Could not retrieve state from DynamoDB", error);
+    }
+    return document;
+  }
+
+  async saveDocumentToDynamo(document, tokens) {
```
beaugrantham-va, line 52: I think this behavior differs slightly from the previous implementation.

There is a chance that the retrieved state could be missing, in which case this query should not run. I remember we had to explicitly add this check but I don't remember the exact circumstances. 

https://github.com/department-of-veterans-affairs/vets-saml-proxy/blob/master/oauth-proxy/oauthHandlers/tokenHandler.js#L82

beaugrantham-va (changes requested):

oauth-proxy/oauthHandlers/tokenHandlerStrategyClasses/tokenHandlerClient.js
```diff
@@ -0,0 +1,137 @@
...
+    this.req = req;
+    this.res = res;
+    this.next = next;
+  }
+  async handleToken() {
+    let clientMetadata;
+    try {
+      clientMetadata = this.createClientMetaData();
+    } catch (error) {
+      if (error.error && error.error == "invalid_client") {
```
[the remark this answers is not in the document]
lukemarkwordtlibertyits: Done.

oauth-proxy/oauthHandlers/tokenHandlerStrategyClasses/tokenHandlerClient.js
```diff
@@ -0,0 +1,137 @@
...
+    const client = new this.issuer.Client(clientMetadata);
+
+    let tokens;
+    try {
+      tokens = await this.tokenHandlerStrategy.getToken(
+        client,
+        this.redirect_uri
+      );
+    } catch (error) {
+      return {
```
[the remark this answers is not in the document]
lukemarkwordtlibertyits: This is consistent with previous functionality.

oauth-proxy/oauthHandlers/tokenHandlerStrategyClasses/tokenHandlerClient.js
lukemarkwordtlibertyits: yes, I took out the status, as it wont be used for the default handler, and you had mentioned it should be 500 earlier.

oauth-proxy/oauthHandlers/tokenHandlerStrategyClasses/refreshTokenStrategy.js
lukemarkwordtlibertyits: I understand your concern, I cannot get redirect_uri working correctly with the okta sdk when it is a member variable. I can take a look.

oauth-proxy/oauthHandlers/tokenHandler.js
```diff
@@ -21,150 +23,61 @@ const tokenHandler = async (
...
+    res,
+    next
+  );
 
-  const client = new issuer.Client(clientMetadata);
+  let tokenResponse;
+  try {
+    tokenResponse = await tokenHandlerClient.handleToken();
+  } catch (err) {
+    req.query.error = err.error;
```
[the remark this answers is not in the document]
lukemarkwordtlibertyits: I can either change the catchall error to take into account non req.body error messages or we could not log this?

oauth-proxy/oauthHandlers/tokenHandlerStrategyClasses/tokenHandlerClient.js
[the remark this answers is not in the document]
lukemarkwordtlibertyits: Nevermind. I see what you mean.

crolarlibertyva:
So far all the retesting passes for me.

oauth-proxy/oauthHandlers/tokenHandlerStrategyClasses/tokenHandlerClient.js
```diff
@@ -0,0 +1,88 @@
...
+          error_description: error.error_description,
+        },
+      };
+    }
+
+    let document = await this.tokenHandlerStrategy.pullDocumentFromDynamo();
+    let state;
+    if (document && tokens) {
+      await this.tokenHandlerStrategy.saveDocumentToDynamo(document, tokens);
+      state = document.state.S;
```
beaugrantham-va, line 50: There is a minor difference here.

The previous implementation explicitly fell back to `state = null` if there was no state.
https://github.com/department-of-veterans-affairs/vets-saml-proxy/blob/master/oauth-proxy/oauthHandlers/tokenHandler.js#L98

This updated implementation potentially results in `state` being `undefined`. The difference is that when it gets added to the response body, an `undefined` value will be absent from the final object `{ }` vs `{ "state": null }`.

oauth-proxy/oauthHandlers/tokenHandlerStrategyClasses/authorizationCodeStrategy.js
```diff
@@ -0,0 +1,74 @@
...
+      rethrowIfRuntimeError(err);
+      this.logger.error("Failed to retrieve document from Dynamo DB.", err);
+    }
+
+    return document;
+  }
+
+  async saveDocumentToDynamo(document, tokens) {
+    try {
+      let state = document.state.S;
```
beaugrantham-va, line 54: Same comment here about a possible runtime exception that should be avoided.

oauth-proxy/oauthHandlers/tokenHandler.js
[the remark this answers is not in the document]
beaugrantham-va: It looks like it still gets logged to a file.

Typically the cause of a 500 error is not displayed to a client, as it could expose runtime details. I think we just skip populating `req.query` and should be good.

beaugrantham-va (changes requested):

beaugrantham-va:
One behavior change worth pointing out.

Previously, a request to `/token` with no credentials AND no `grant_type` would result in a 401 error.
```
{"error": "invalid_client", "error_description": "Client authentication failed"}
```

The new behavior is a 400 error.
```
{"error":"unsupported_grant_type","error_description":"Only authorization and refresh_token grant types are supported"}
```

I think both are equally acceptable. Just documenting it here.

beaugrantham-va (approved):

dbrown-va (approved):

PR #167, opened 2020-10-28
API-2333 - OIDC timeouts

beaugrantham-va:
Setting a custom timeout changed with v 3.0.0 of oidc-client (https://github.com/panva/node-openid-client/tree/v3.15.10/docs#customizing).

Verified timeout is applied by setting it to a low value (400ms) and seeing that a refresh token request occasionally times-out at the given limit.

https://vajira.max.gov/browse/API-3222

### DIFF
--- a/oauth-proxy/index.js
+++ b/oauth-proxy/index.js
@@ -57,7 +57,6 @@ async function createIssuer(upstream_issuer, upstream_issuer_timeout_ms) {
       options.timeout = upstream_issuer_timeout_ms;
       return options;
     };
-
   }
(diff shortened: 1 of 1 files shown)


dbrown-va (approved):
Confirmed changes against the node openid client documentation for default timeout options.

https://github.com/panva/node-openid-client/blob/master/docs/README.md#customizing-http-requests

crolarlibertyva (approved):
Confirmed timeout effect with setting parameter as well as well as maintaining behaviour in default, settings.

PR #170, opened 2020-10-30
API-1852: Support saml-proxy ECS in Staging

mramirez-va:
We've added support for generating the config.json file required by the saml-proxy app. The dockerfile for the deployment into Fargate is being updated to exercise this change.

https://vajira.max.gov/browse/API-1852

Related: #140


beaugrantham-va (approved):

duganth-va (approved):
⭐ Excited this is starting 👍

PR #171, opened 2020-10-30
API-3296 Removing unnecessary error logging.

bdgrider:


oauth-proxy/oauthHandlers/tokenHandlerStrategyClasses/tokenHandlerClient.js
```diff
@@ -79,7 +79,6 @@ class TokenHandlerClient {
         error_description:
           "Could not find a valid patient identifier for the provided authorization code.",
       };
-      this.logger.error(returnError.error_description, error);
```
crolarlibertyva, line 82: I'll have to get a test run in to see what's gonna show in a message.

beaugrantham-va, line 82: _Some_ amount of logging is necessary. Otherwise we will never be able to track down the cause of a problem.

beaugrantham-va (changes requested):

oauth-proxy/oauthHandlers/tokenHandlerStrategyClasses/tokenHandlerClient.js
beaugrantham-va: https://github.com/axios/axios#handling-errors

beaugrantham-va (approved):

crolarlibertyva (commented):
Confirmed the error message shows, but not the token.

crolarlibertyva (approved):

PR #172, opened 2020-11-02
Api 3017 refactor dynamo client to support multiple tables

lukemarkwordtlibertyits:
In order to save launch Ids and access tokens in their own dynamo table, the dynamo_client had to be refactored to support multiple tables. 

I also saw this as an opportunity to separate the dynamo methods in the tokenHandler strategies into their own classes, to better fit the strategy pattern and eliminate the need to reuse code.


lukemarkwordtlibertyits:
Opinions on having each strategy class in its own module. Would it be better to stick to this practice (high file count, low code per file), or should each strategy category be represented in one module that contains all the the strategy classes (low file count, high code per module)? I personally am unsure which practice is better.

beaugrantham-va:
I'm having a hard time picking out what changed in this PR vs the other commits (like #173 and #168). Not sure why these are showing up as changes.

PR #173, opened 2020-11-02
Api 3015 oauth proxy client credentials flow

dbrown-va:
OAuth Proxy functionality added to support Client Credentials flow for configured trusted consumers.

Only JWT Bearer client_assertion's are valid.  The ICN can be supplied as an additional body arg as `launch`.

JIRA Issue: https://vajira.max.gov/browse/API-3015

### DIFF
--- a/oauth-proxy/oauthHandlers/tokenHandler.js
+++ b/oauth-proxy/oauthHandlers/tokenHandler.js
@@ -151,7 +151,7 @@ const getTokenStrategy = (
     ) {
       throw {
         status: 400,
-        error: "Client assertion error",
+        error: "invalid_request",
         error_description: "Client assertion type must be jwt-bearer.",
       };
     }
--- a/oauth-proxy/oauthHandlers/tokenHandler.js
+++ b/oauth-proxy/oauthHandlers/tokenHandler.js
@@ -6,6 +6,9 @@ const {
   AuthorizationCodeStrategy,
 } = require("./tokenHandlerStrategyClasses/authorizationCodeStrategy");
 const {
+  ClientCredentialsStrategy,
+} = require("./tokenHandlerStrategyClasses/clientCredentialsStrategy");
+const {
   UnsupportedGrantStrategy,
 } = require("./tokenHandlerStrategyClasses/unsupportedGrantStrategy");
 const {
@@ -86,6 +89,8 @@ function createClientMetadata(redirect_uri, req, config) {
     clientMetadata.token_endpoint_auth_method = "none";
     clientMetadata.client_id = req.body.client_id;
     delete req.body.client_id;
+  } else if (req.body.grant_type === "client_credentials") {
+    //Unused by client credentials flow
   } else {
     throw {
       error: "invalid_client",
@@ -136,6 +141,14 @@ const getTokenStrategy = (
       dynamoClient,
       redirect_uri,
       getClient(issuer, redirect_uri, req, config)
+    );
+  } else if (req.body.grant_type === "client_credentials") {
+    tokenHandlerStrategy = new ClientCredentialsStrategy(
+      req,
+      logger,
+      dynamo,
+      dynamoClient,
+      issuer.token_endpoint
     );
   } else {
     tokenHandlerStrategy = new UnsupportedGrantStrategy();
--- a/oauth-proxy/oauthHandlers/tokenHandler.js
+++ b/oauth-proxy/oauthHandlers/tokenHandler.js
@@ -90,8 +90,6 @@ function createClientMetadata(redirect_uri, req, config) {
     clientMetadata.token_endpoint_auth_method = "none";
     clientMetadata.client_id = req.body.client_id;
     delete req.body.client_id;
-  } else if (req.body.grant_type === "client_credentials") {
-    //Unused by client credentials flow
   } else {
     throw {
       error: "invalid_client",
--- a/oauth-proxy/oauthHandlers/tokenHandler.js
+++ b/oauth-proxy/oauthHandlers/tokenHandler.js
@@ -145,6 +145,13 @@ const getTokenStrategy = (
       validateToken
     );
   } else if (req.body.grant_type === "client_credentials") {
+    if (req.body.client_assertion_type !== "urn:ietf:params:oauth:client-assertion-type:jwt-bearer") {
+      throw {
+        status: 400,
+        error: "Client assertion error",
+        error_description: "Client assertion type must be jwt-bearer.",
+      };
+    }
     tokenHandlerStrategy = new ClientCredentialsStrategy(
       req,
       logger,
--- a/oauth-proxy/oauthHandlers/tokenHandler.js
+++ b/oauth-proxy/oauthHandlers/tokenHandler.js
@@ -36,7 +36,8 @@ const tokenHandler = async (
       dynamo,
       dynamoClient,
       config,
-      req
+      req,
+      validateToken
     );
   } catch (error) {
     rethrowIfRuntimeError(error);
@@ -122,7 +123,8 @@ const getTokenStrategy = (
   dynamo,
   dynamoClient,
   config,
-  req
+  req,
+  validateToken
 ) => {
   let tokenHandlerStrategy;
   if (req.body.grant_type === "refresh_token") {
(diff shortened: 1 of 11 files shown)


oauth-proxy/oauthHandlers/tokenHandler.js
```diff
@@ -135,7 +141,26 @@ const getTokenStrategy = (
...
+      validateToken
+    );
+  } else if (req.body.grant_type === "client_credentials") {
+    if (
+      req.body.client_assertion_type !==
+      "urn:ietf:params:oauth:client-assertion-type:jwt-bearer"
+    ) {
+      throw {
+        status: 400,
+        error: "Client assertion error",
```
beaugrantham-va, line 154: There are a fixed set of `error` values that should be returned per the RFC.

https://tools.ietf.org/html/rfc6749#section-5.2

I think this falls into the category of `invalid_client` and should actually return a 401.

oauth-proxy/oauthHandlers/tokenHandlerStrategyClasses/clientCredentialsStrategy.js
```diff
@@ -0,0 +1,85 @@
...
+        }
+      } else if (error.response.status == 401) {
+        throw {
+          statusCode: 401,
+          error: error.response.data.error,
+          error_description: error.response.data.error_description,
+        };
+      } else {
+        throw {
+          statusCode: 500,
```
beaugrantham-va, line 59: Will this get handled by Sentry? All our other 500s just bubble up to the "unknown error occurred" handler I think.

Also we should include some level of logging here so we have an idea of what the actual problem is.

oauth-proxy/oauthHandlers/tokenHandlerStrategyClasses/clientCredentialsStrategy.js
```diff
@@ -0,0 +1,85 @@
...
+class ClientCredentialsStrategy {
+  constructor(req, logger, dynamo, dynamoClient, token_endpoint) {
+    this.req = req;
+    this.logger = logger;
+    this.dynamo = dynamo;
+    this.dynamoClient = dynamoClient;
+    this.token_endpoint = token_endpoint;
+  }
+
+  //will throw error if cannot retrieve refresh token
```
beaugrantham-va, line 13: This comment was copied from somewhere and can go away.

oauth-proxy/oauthHandlers/tokenHandlerStrategyClasses/clientCredentialsStrategy.js
lukemarkwordtlibertyits: This error will get handled. It may be useful to put the `rethrowIfRuntimeError` method at the top of this handler.

oauth-proxy/oauthHandlers/tokenHandlerStrategyClasses/clientCredentialsStrategy.js
```diff
@@ -0,0 +1,85 @@
...
+        token = res.data;
+      } else {
+        throw {
+          statusCode: 500,
+          error: "token_failure",
+          error_description: "Failed to retrieve access_token.",
+        };
+      }
+    } catch (error) {
+      if (error.response.status == 400) {
```
lukemarkwordtlibertyits, line 37: `rethrowIfRuntimeError` could go here

oauth-proxy/oauthHandlers/tokenHandlerStrategyClasses/clientCredentialsStrategy.js
```diff
@@ -0,0 +1,85 @@
...
+        if (error.response.data.errorCode) {
+          throw {
+            statusCode: 400,
+            error: error.response.data.errorCode,
+            error_description: error.response.data.errorSummary,
+          };
+        } else {
+          throw {
+            statusCode: 400,
```
lukemarkwordtlibertyits, line 46: confirmed this is working properly.

oauth-proxy/oauthHandlers/tokenHandlerStrategyClasses/clientCredentialsStrategy.js
```diff
@@ -0,0 +1,85 @@
...
+            error: error.response.data.error,
+            error_description: error.response.data.error_description,
+          };
+        }
+      } else if (error.response.status == 401) {
+        throw {
+          statusCode: 401,
```
lukemarkwordtlibertyits, line 53: Confirmed this is working properly.

lukemarkwordtlibertyits (commented):
Everything seems to be working so far. I will also have to check out the tests when the get finished.

oauth-proxy/oauthHandlers/tokenHandlerStrategyClasses/clientCredentialsStrategy.js
lukemarkwordtlibertyits: before 400 check that is

beaugrantham-va:
If I set `-H "Content-type: application/json"` I get a 500 error. I would expect this to be a 400 indicating a problem with the request.

oauth-proxy/tests/clientCredentialsStrategy.test.js
```diff
@@ -0,0 +1,126 @@
+require("jest");
```
lukemarkwordtlibertyits, line 1: I like that you are testing the strategy itself. This is one of the advanteges of the strategy pattern. You can test functionality on a smaller scale. This will lead to more coverage, with less tests. 

The future of the token handler tests will probably resemble more of this testing pattern than testing the token handler and its strategies.

oauth-proxy/tests/clientCredentialsStrategy.test.js
```diff
@@ -0,0 +1,126 @@
...
+      dynamoClient,
+      token_endpoint
+    );
+
+    let token = await clientCredentialsStrategy.getTokenResponse();
+    expect(token).toEqual(data);
+  });
+
+  it("handles invalid client_credentials request", async () => {
+    let req = new MockExpressRequest({
```
lukemarkwordtlibertyits, line 65: Currently the code coverage for the clientCredentialStrategy is at 79 percent. We will likely bump up the coverage metrics soon. Duplicating this test to return a 400 instead of a 401 will bump up the clientCredentialStrategy code coverage to 87%.

beaugrantham-va:
> If I set `-H "Content-type: application/json"` I get a 500 error. I would expect this to be a 400 indicating a problem with the request.

Confirmed this is an existing bug which was not introduced by this branch. Created https://vajira.max.gov/browse/API-3387 to address this.

oauth-proxy/tests/clientCredentialsStrategy.test.js
dbrown-va: Agreed here.  Added both 400 and 500 tests in addition to the 401.  That gets the coverage up to 89.66

oauth-proxy/oauthHandlers/tokenHandlerStrategyClasses/clientCredentialsStrategy.js
```diff
@@ -0,0 +1,93 @@
...
+        method: "post",
+        url: this.token_endpoint,
+        data: data,
+        headers: this.req.headers,
+      });
+      if (res.status == 200) {
+        token = res.data;
+      } else {
+        throw {
+          statusCode: 500,
```
beaugrantham-va, line 31: The same error logging needs to go here too.

oauth-proxy/oauthHandlers/tokenHandlerStrategyClasses/clientCredentialsStrategy.js
```diff
@@ -0,0 +1,93 @@
...
+      res = await axios({
+        method: "post",
+        url: this.token_endpoint,
+        data: data,
+        headers: this.req.headers,
+      });
+      if (res.status == 200) {
+        token = res.data;
+      } else {
+        throw {
```
beaugrantham-va, line 30: There needs to be some logging here.

beaugrantham-va: Or should this just be throwing the error and letting the catch decide how to handle and log all errors?

beaugrantham-va (approved):

oauth-proxy/oauthHandlers/tokenHandlerStrategyClasses/clientCredentialsStrategy.js
dbrown-va: Added logging including status code.

PR #176, opened 2020-11-12
Api 3381 token handler 500 errors to sentry

lukemarkwordtlibertyits:
Validation errors will now be sent to sentry.

### DIFF
--- a/oauth-proxy/oauthHandlers/tokenHandlerStrategyClasses/getPatientInfoStrategies/getPatientInfoFromValidateEndpointStrategy.js
+++ b/oauth-proxy/oauthHandlers/tokenHandlerStrategyClasses/getPatientInfoStrategies/getPatientInfoFromValidateEndpointStrategy.js
@@ -25,6 +25,7 @@ class GetPatientInfoFromValidateEndpointStrategy {
         error: "invalid_grant",
         error_description:
           "Could not find a valid patient identifier for the provided authorization code.",
+        status: 500
       };
     }
     return patient;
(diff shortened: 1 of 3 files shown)


oauth-proxy/oauthHandlers/tokenHandlerStrategyClasses/getPatientInfoStrategies/getPatientInfoFromValidateEndpointStrategy.js
```diff
@@ -25,6 +25,7 @@ class GetPatientInfoFromValidateEndpointStrategy {
         error: "invalid_grant",
         error_description:
           "Could not find a valid patient identifier for the provided authorization code.",
+        status: 500,
```
lukemarkwordtlibertyits, line 28: Question: Should the sentry handler function account for null statuses? Currently, Sentry handler will not report errors with null statuses.

beaugrantham-va: > Question: Should the sentry handler function account for null statuses? Currently, Sentry handler will not report errors with null statuses.

Anything that bubbles up to the catch-all 500 error handler should be handled by Sentry. I thought that this handler would catch anything that had a null status.

lukemarkwordtlibertyits: The current code is `return status >= 400` so if its null it wont send to sentry. It would be an easy fix though.

beaugrantham-va (approved):

PR #178, opened 2020-11-16
API-3309 - Cache session index

bdgrider:
https://vajira.max.gov/browse/API-3309


saml-proxy/docker-compose.yml
```diff
@@ -1,9 +1,15 @@
 version: '2.1'
 services:
+  redis:
+    image: redis
+    networks: 
+      - 'default'
+    expose:
```
crolarlibertyva, line 7: port mapping localhost eg
`    ports:
      - 6379:6379
`

crolarlibertyva (changes requested):

crolarlibertyva (approved):

saml-proxy/src/routes/index.js
```diff
@@ -168,7 +170,9 @@ export default function configureExpress(
     }
   });
 
-  addRoutes(app, idpOptions, spOptions);
+  /* exported variableName */
+  //const cache = new RedisCache();
```
esubocz-va, line 174: this comment can probably be taken out?

esubocz-va (approved):

saml-proxy/src/routes/acsHandlers.ts
```diff
@@ -168,6 +168,45 @@ export const testLevelOfAssuranceOrRedirect = (
   }
 };
 
+export const validateIdpResponse = (cache: ICache) => {
+  return async (req: IConfiguredRequest, res: Response, next: NextFunction) => {
+    const sessionIndex = req?.user?.authnContext?.sessionIndex;
```
dbrown-va, line 173: Out of an abundance of caution, I would lean towards adding a flag for us to be able to enable solely in dev for a couple days to ensure expected behavior prior to enabling in higher environments.

We will also need a corresponding devops PR anyway once the cache set up is in place.

dbrown-va (approved):
Confirmed locally with the `cacheEnabled` behavior both true & false.

Devops PR already merged and only enables dev & staging at this time.

PR #181, opened 2020-11-24
Api 3601 dynamo ttl

crolarlibertyva:
- Began the adoption of the Dynamo DB Document Client API as opposed to the current usage of the lower-level Database API.
- Also, using this to include the access token expiration as a field with the intent to use it as the "time to live" attribute for the record in the client credentials table.


dbrown-va (changes requested):
We want the `exp` from the token itself, rather than the `client_assertion`, which will avoid passing the `assert_info`.

I think we can avoid modifying the client credentials strategy as well.

oauth-proxy/dynamo_client.js
```diff
@@ -4,15 +4,26 @@
...
+    dynamoDb = new AWS.DynamoDB({ endpoint: `http://${local}` });
+    AWS.config.update({
+      region: "us-west-2",
+      endpoint: `http://${local}`,
+    });
+    dynamoDb.dbDocClient = new AWS.DynamoDB.DocumentClient();
+  } else {
+    AWS.config.update(awsConfig);
+    dynamoDb = new AWS.DynamoDB();
+    AWS.config.update(awsConfig);
```
beaugrantham-va, line 22: Why is this same update statement run twice?

oauth-proxy/dynamo_client.js
```diff
@@ -4,15 +4,26 @@
...
+    dynamoDb.dbDocClient = new AWS.DynamoDB.DocumentClient();
   }
-  return new DynamoDB();
+
+  return dynamoDb;
```
beaugrantham-va, line 26: How does this change affect strategies that read/write which were not updated in this PR?

oauth-proxy/dynamo_client.js
```diff
@@ -145,11 +156,31 @@ function saveToDynamoAccessToken(client, accessToken, key, value, TableName) {
   });
 }
 
+function savePayloadToDynamo(dynamoDb, payload, tableName) {
```
beaugrantham-va, line 159: Should this function be async since it returns a promise?

oauth-proxy/dynamo_client.js
crolarlibertyva: Without the 2nd update, it didn't retain the config settings on the subsequent client substantiation.

oauth-proxy/dynamo_client.js
crolarlibertyva: This leaves the original read/writes intact, just needed to add a reference to the document client for use with the expire_on.

beaugrantham-va: Should new code be using `savePayloadToDynamo` and the old functions deprecated?

crolarlibertyva: Yes, will followup with a refactor ticket to address the comment of the module as well.

beaugrantham-va (approved):

dbrown-va (approved):

dbrown-va:
Confirmed update to use the expiration from the token (`decodedToken.exp`) itself rather the client assertion.

PR #188, opened 2020-12-08
SAML Proxy DockerFG use EFS mount path

mramirez-va:
https://vajira.max.gov/browse/API-3865
- set docker labels
- update config file path to EFS mount path


beaugrantham-va (approved):

duganth-va (approved):
Sweet!

PR #190, opened 2020-12-09
Api 3493 filter logs

crolarlibertyva:
Minimal error logging to keep sensitive information out of the logs.


oauth-proxy/index.js
```diff
@@ -322,8 +323,7 @@ function buildApp(
   }
 
   app.use(function (err, req, res, next) {
-    logger.error(err);
-
+    logger.error(minimalError(err));
```
beaugrantham-va, line 326: I'm concerned that this impacts more than axios errors. 

All un-handled errors end up here. We can't be sure of the structure of errors from other libs and may want to include more detail.

If I `throw { detail: "This detail is important" };` then I end up with this:

```
oauth-proxy_1         | {
oauth-proxy_1         |   "message": {},
oauth-proxy_1         |   "level": "error",
oauth-proxy_1         |   "service": "oauth-proxy",
oauth-proxy_1         |   "time": "2020-12-09T14:35:04.932Z"
oauth-proxy_1         | }
```

I'd prefer this be constrained to just the axios calls if possible.

crolarlibertyva: Roger that.

beaugrantham-va: A quick look through the API docs landed me at interceptors:

https://github.com/axios/axios#interceptors

This might be worth exploring. The error object could be filtered/re-mapped here and then treated as any other generic error. 

We'd have to make sure that errors we proxy directly to the client still work as expected.

oauth-proxy/oauthHandlers/authorizeHandler.js
```diff
@@ -42,7 +42,7 @@ const authorizeHandler = async (
   } catch (error) {
     logger.error(
       "Unrecoverable error: could not get the Okta client app",
-      error
+      minimalError(error)
```
dbrown-va, line 45: Ben's PR for API-3487 is addressing this error scenario.  I recommend reverting this change and focusing on the token calls to avoid conflicts.

bdgrider (approved):
All unit tests and regression tests are passing. Can do a re-run after you address comments.

oauth-proxy/index.js
beaugrantham-va: Update after talking with Larry - 

The specific log entries in question for the /token endpoint are actually from the oidc client (not from axios).

The interceptors and axios docs I referenced above do not apply here.

The failed axios calls _do_ need to be addressed but may be handled with a future PR.

beaugrantham-va (approved):

oauth-proxy/utils.js
```diff
@@ -81,6 +81,36 @@ const hashString = (unhashedString, secret) => {
   return hashedString;
 };
 
+const isString = (error) =>
```
dbrown-va, line 84: Not a big deal, but this new `isString` method isn't really specific to being an error, something like `obj` or `myVar` could be a better variable name.

oauth-proxy/tests/utils.test.js
```diff
@@ -168,3 +169,100 @@ describe("parseBearerAuthorization", () => {
     expect(parseBearerAuthorization("Bearer jwt")).toBe("jwt");
   });
 });
+
+describe("minimalError", () => {
+  it("API-3493 verbose log on timeout fix verification", () => {
+    let testError = {
```
dbrown-va, line 175: I think we can trim down this `testError` significantly while still validating the functionality.  The filtering is inconsistent with several places removing the authServer ID, but others still including it.  Such that it may cause confusion.

crolarlibertyva: Roger that.

PR #191, opened 2020-12-09
API-3487 Adding graceful handling change plus unit test.

bdgrider:


oauth-proxy/oauthHandlers/authorizeHandler.js
```diff
@@ -40,13 +40,19 @@ const authorizeHandler = async (
...
     logger.error(
       "Unrecoverable error: could not get the Okta client app",
       error
     );
-    // This error is unrecoverable because we would be unable to verify
-    // that we are redirecting to a whitelisted client url
-    return next(error);
+
+    res.status(400).json({
+      error: error.name,
```
beaugrantham-va, line 51: The value of `error.name` here will be something like `OktaApiError`. The OAuth RFC specifies a list of valid `error` values and we need to stick to those. I think the appropriate value here is `invalid_client` (like above if a redirect URL is invalid).

I would also recommend a static `error_description` as not to expose any detail about the back-end system. We can't guarantee what that message will be now and in the future.

```
    res.status(400).json({
      error: "invalid_client",
      error_description: "The client specified by the application is not valid."
    });
```

dbrown-va (approved):
Confirmed the syntax change suggested earlier has been applied to match the OAuth RFC.

beaugrantham-va (approved):

PR #193, opened 2020-12-10
Api 3387 correct 400 message

crolarlibertyva:
Addresses incorrect error reporting on processing the request.

https://vajira.max.gov/browse/API-3387


beaugrantham-va:
Where is the origin / throw point for the error in the story?

I tend to think this scenario should be handled explicitly with validation vs falling into the un-handled error bucket. We should be ensuring the input is of the type we expect (string vs object / form vs json) instead of assuming and relying on a catch-all.

The response also needs to be a valid oauth response. Something like below.

```
{
  "error": "invalid_request",
  "error_description": "Invalid or unsupported content-type"
}
```

crolarlibertyva:
> Where is the origin / throw point for the error in the story?
> 
> I tend to think this scenario should be handled explicitly with validation vs falling into the un-handled error bucket. We should be ensuring the input is of the type we expect (string vs object / form vs json) instead of assuming and relying on a catch-all.
> 
> The response also needs to be a valid oauth response. Something like below.
> 
> ```
> {
>   "error": "invalid_request",
>   "error_description": "Invalid or unsupported content-type"
> }
> ```
The error originates up the chain and our catch-all error is the only spot I've been able to latch on to it.
"SyntaxError: Unexpected token g in JSON at position 0\n    at JSON.parse (<anonymous>)\n    at createStrictSyntaxError (/home/node/node_modules/body-parser/lib/types/json.js:158:10)\n    at parse (/home/node/node_modules/body-parser/lib/types/json.js:83:15)\n    at /home/node/node_modules/body-parser/lib/read.js:121:18\n    at invokeCallback (/home/node/node_modules/raw-body/index.js:224:16)\n    at done (/home/node/node_modules/raw-body/index.js:213:7)\n    at IncomingMessage.onEnd (/home/node/node_modules/raw-body/index.js:273:7)\n    at IncomingMessage.emit (events.js:314:20)\n    at IncomingMessage.EventEmitter.emit (domain.js:483:12)\n    at endReadableNT (_stream_readable.js:1241:12)"

oauth-proxy/index.js
```diff
@@ -327,11 +327,29 @@ function buildApp(
     // If we have error and description as query params display them, otherwise go to the
     // catchall error handler
     const { error, error_description } = req.query;
-    if (error && error_description) {
-      res.status(500).send(`${error}: ${error_description}`);
+    let statusCode = 500;
```
dbrown-va, line 330: For the 500 errors, per the OAuth RFC these really should be fixed "server_error" responses, which is handled in the API-3329 PR (https://github.com/department-of-veterans-affairs/vets-saml-proxy/pull/194)

https://vajira.max.gov/browse/API-3329?focusedCommentId=760812&page=com.atlassian.jira.plugin.system.issuetabpanels:comment-tabpanel#comment-760812

dbrown-va: For simplicity, I think we merge the API-3329 PR first, and then resolve the conflicts here.

oauth-proxy/index.js
```diff
@@ -326,17 +326,26 @@ function buildApp(
...
+    let error_out = {
+      error: "sever_error",
+    };
+    if (err.statusCode && err.statusCode === 400) {
+      statusCode = err.statusCode;
     }
+    if (statusCode === 400) {
+      error_out = {
+        error: "invalid_request",
+        error_description: "Invalid or unsupported content-type",
```
beaugrantham-va, line 340: I'm still concerned that this is not the right place to be catching this error.

I also don't think it's safe to assume that if any 400 error makes it this far that it must be the mismatched content-type error.

I'm requesting changes on this PR but don't have a recommended change. I just want to have a chance to look deeper into this before it gets merged.

beaugrantham-va (changes requested):

oauth-proxy/index.js
beaugrantham-va: I little digging later and I have a recommendation.

The error originates from the `body-parser` lib. We have the following in index.js which attempts to parse the body based on content-type and populate the `req` object.

```
const bodyParser = require("body-parser");
app.use(bodyParser.json({ extended: true }));
```

If the content-type is application/json and the body is not, the lib throws the error that we're seeing. This occurs before any of our oauth / token handlers are called.

```
oauth-proxy_1         | {
oauth-proxy_1         |   "expose": true,
oauth-proxy_1         |   "statusCode": 400,
oauth-proxy_1         |   "status": 400,
oauth-proxy_1         |   "body": "...",
oauth-proxy_1         |   "type": "entity.parse.failed",
oauth-proxy_1         |   "level": "error",
oauth-proxy_1         |   "service": "oauth-proxy",
oauth-proxy_1         |   "time": "2020-12-13T14:45:27.388Z"
oauth-proxy_1         | }
```

I'm not a fan of libs throwing HTTP status errors. This is not their responsibility. I dug around their GitHub project and I see multiple issues reported around the same topic, and requests for graceful handling of this specific error. They all seem to be closed without resolution.

The general recommendations seem to point to a similar solution that you have - handling this in a catch-all. I think this should be explicit though, as not to group all random lib-thrown 400 errors into the same bucket.

```
if (err.type == "entity.parse.failed") {
  res.status(400).send({ error: "invalid_request", error_description: "something about content-type" });
} else {
  // existing 500 error logic
}
```

oauth-proxy/index.js
```diff
@@ -326,17 +328,28 @@ function buildApp(
...
-        error_description: error_description,
-      });
-    } else {
-      res.status(500).json({
-        error: "server_error",
-      });
+    const { error_description } = req.query;
+    let statusCode = 500;
+    let error_out = {
+      error: "sever_error",
```
dbrown-va, line 334: Just a spelling issue here, this should be `se**r**ver_error`

oauth-proxy/index.js
```diff
@@ -326,17 +328,28 @@ function buildApp(
...
+      err.statusCode &&
+      err.statusCode === 400 &&
+      err.type === "entity.parse.failed"
+    ) {
+      statusCode = 400;
+      error_out = {
+        error: "invalid_request",
+        error_description: "Invalid or unparsable content-type",
+      };
+    } else if (isString(err)) {
```
beaugrantham-va, line 346: This 500 behavior is different. This should be a like-for-like replacement with the addition of the new 400 error.

If we don't already have a unit test (not a integration / e2e / axios test) exercising these branches please create one.

beaugrantham-va (changes requested):

beaugrantham-va (approved):

dbrown-va (approved):

PR #194, opened 2020-12-10
API-3329 Adding all changes to make 500 errors use index.js handler w…

bdgrider:
Adding all changes to make 500 errors use index.js handler with oauth2 error from RFC.

### DIFF
--- a/oauth-proxy/index.js
+++ b/oauth-proxy/index.js
@@ -328,9 +328,14 @@ function buildApp(
     // catchall error handler
     const { error, error_description } = req.query;
     if (error && error_description) {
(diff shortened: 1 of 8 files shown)


oauth-proxy/index.js
```diff
@@ -328,9 +328,14 @@ function buildApp(
     // catchall error handler
     const { error, error_description } = req.query;
     if (error && error_description) {
```
crolarlibertyva, line 330: In the PR I submitted earlier, you may run into a conflict in this section.

bdgrider:
image

example of a 500 server error

dbrown-va (approved):

PR #196, opened 2020-12-11
Api 3328 oauth regression script small files

lukemarkwordtlibertyits:
e2e regression tests, broken into smaller files. 

## Quick Start

```
export USER_EMAIL=va.api.user+idme.001@gmail.com
export USER_PASSWORD=Password1234!
export CLIENT_ID={{ client id }}
export CLIENT_SECRET={{ client secret }}
export HOST=https://sandbox-api.va.gov/oauth2
export CC_CLIENT_ID={{ client id }}
export CC_CLIENT_SECRET={{ client secret }}
```

```
./tests/scripts/regression_tests.sh
```


esubocz-va:
it might be helpful to have usage methods in the scripts, like in entrypoint scripts. ex: https://github.com/department-of-veterans-affairs/lighthouse-auth/blob/66f9a8f7ea85c7e075c7e82747f9e5b3906be2e5/auth-tests/src/main/docker/entrypoint.sh#L15

crolarlibertyva:
So far so good with getting the effects we want with a happy path and a timeout failure.  But I do get this on the happy path test.:
`jq: error: Could not open file /tmp/tmp.LLe0OB8ii3: No such file or directory
`
This doesn't deter the testing from passing when you expect it to pass and fail when you expect it to fail.

lukemarkwordtlibertyits:
> So far so good with getting the effects we want with a happy path and a timeout failure. But I do get this on the happy path test.:
> `jq: error: Could not open file /tmp/tmp.LLe0OB8ii3: No such file or directory `
> This doesn't deter the testing from passing when you expect it to pass and fail when you expect it to fail.

@crolarlibertyva can you send me the steps to replicate?

The tests use temporary files to share variables across scripts. One of the disadvantages to breaking out the tests into smaller scripts. Most likely the machine did not create a temp file for one of the variables. If I could replicate this. I could create a check for this file.

oauth-proxy/tests/scripts/introspect_tests.sh
```diff
@@ -0,0 +1,187 @@
...
+pass=1
+curl_status="$(mktemp)"
+curl_body="$(mktemp)"
+
+usage() {
+cat <<EOF
+Tests the Oauth Proxy's Introspect endpoint.
+
+Example
+  ./introspect_tests.sh --host=https://sandbox-api.va.gov/oauth2 --tokens={ Tokens } --expired-access={ expired access token }
```
dbrown-va, line 18: I think the maintenance of temp_files adds complexity and may cause issues unrelated to the flow itself.

Instead of parsing like `access_token=$(echo "$TOKENS" | jq ".access_token" | tr -d '"')`, we would be better off passing any required arguments to each individual script.  This makes running individual scripts easier than having to build up the expected file structure.

dbrown-va:
> > So far so good with getting the effects we want with a happy path and a timeout failure. But I do get this on the happy path test.:
> > `jq: error: Could not open file /tmp/tmp.LLe0OB8ii3: No such file or directory `
> > This doesn't deter the testing from passing when you expect it to pass and fail when you expect it to fail.
> 
> @crolarlibertyva can you send me the steps to replicate?
> 
> The tests use temporary files to share variables across scripts. One of the disadvantages to breaking out the tests into smaller scripts. Most likely the machine did not create a temp file for one of the variables. If I could replicate this. I could create a check for this file.

Rather than using temp files, I think we would be better off passing required inputs individually to each sub-script.  That makes them easier to use independently vs having to build up a file or files to exercise the calls.



```
access_token=$(echo "$TOKENS" | jq ".access_token" | tr -d '"')
```

saml-proxy/src/VetsAPIClient.ts
```diff
@@ -45,7 +45,8 @@ export class VetsAPIClient {
       gender: user.gender || null,
       level_of_assurance: "3",
     };
-
+    console.log(`${this.apiHost}${MVI_PATH}`);
+    console.log(JSON.stringify(body));
```
beaugrantham-va, line 49: What is this?

dbrown-va:
Accidently closed.  Meant to cancel my comment as I did it inline instead.  Apologies.

beaugrantham-va (dismissed):

saml-proxy/src/VetsAPIClient.ts
lukemarkwordtlibertyits: my bad

oauth-proxy/tests/scripts/introspect_tests.sh
```diff
@@ -0,0 +1,187 @@
+#!/usr/bin/env bash
+# Team Pivot!
+
+HOST=
+TOKENS=$2
+EXPIRED_ACCESS=$3
+DIR="$( cd "$( dirname "${BASH_SOURCE[0]}" )" >/dev/null 2>&1 && pwd )"
```
beaugrantham-va, line 7: A slightly cleaner option is `$(readlink -f $(dirname $0))`

oauth-proxy/tests/scripts/introspect_tests.sh
lukemarkwordtlibertyits: Temp files are used a couple places in the script. I think for the `curl_body` and `curl_status` variable they work fine. Sharing variables between scripts is necessary to reduce auth bot calls (in particular the tokens generated in token_tests are used in the introspect tests). The auth bot causes the most false negatives and is costly to run, so it should only be used once to increase speed and accuracy. 

Bash variables cannot be shared from sub shell to parent shell. The TOKEN and EXPIRED_ACCESS temp files were created to solve this problem. Allowing subscripts to use the tokens created in the token_tests script. I originally did not split the tests into multiple scripts to avoid this; however, when Beau suggested splitting the scripts I decided to put this PR out, so we could weigh out the alternatives. 

Currently I have gotten around this issue, by forcing the token_test script to be ran in the regression_test shell and exporting the TOKEN and EXPIRED_ACCESS variables. Once the temp files are removed, I am not sure that jq parsing will add unexpected issues to the script, but I could do the parsing in the regression_tests script if you would perrfer.

bdgrider (approved):
Looks good to me, all tests passing in my environment.

oauth-proxy/tests/scripts/introspect_tests.sh
```diff
@@ -0,0 +1,187 @@
...
+    -o "$curl_body" \
+    -d "token_type_hint=$hint" \
+    -d "token=$token" \
+    -d "client_id=$client_id" \
+    -d "client_secret=$client_secret" \
+    "$HOST/introspect" > "$curl_status"
+}
+
+# -------
+echo -e "\nIntropsect Tests"
```
dbrown-va, line 91: Just a spelling issue here.

oauth-proxy/tests/scripts/introspect_tests.sh
beaugrantham-va: > in particular the tokens generated in token_tests are used in the introspect tests

Long term, I think all of the broken-out tests should be independent. If introspect needs a token to test something, it should get its own token. I get that we might not be ready for that if the oauth-bot is problematic.

The other option for sharing functions and vars between scripts is to break things up and `source` them. Doing this should allow a script could call `do_assert abc def` (or whatever the function name is) instead of executing `./assert.sh --arg1 abc --arg2 def`. 

In general, I think this PR satisfies "providing a way to regression test the application as a whole, with manageable sized scripts".

I think there will be continual refining of these (both in terms of content and structure) as they start being utilized.

dbrown-va (approved):

PR #201, opened 2020-12-15
expect_json_body compares by variables now

lukemarkwordtlibertyits:
Eliminated the use of temp files in the expect_json_body method. Now the regression tests are compatible with versions of jq under 1.6.

### DIFF
--- a/oauth-proxy/tests/scripts/assertions.sh
+++ b/oauth-proxy/tests/scripts/assertions.sh
@@ -100,16 +100,11 @@ expect_status() {
 # the JSON in file $curl_body
 #
 expect_json_body() {
-  expected_body="$(mktemp)"
-  echo "$EXPECTED_JSON" > "$expected_body"
-  actual_body="$(mktemp)"
-  echo "$JSON" > "$actual_body"
-
-  if [[ -z "$JSON" ]] || [ "$(cmp <(jq -cS . "$actual_body") <(jq -cS . "$expected_body"))" ]; then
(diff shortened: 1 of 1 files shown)


oauth-proxy/tests/scripts/assertions.sh
```diff
@@ -100,16 +100,11 @@ expect_status() {
...
-  actual_body="$(mktemp)"
-  echo "$JSON" > "$actual_body"
-
-  if [[ -z "$JSON" ]] || [ "$(cmp <(jq -cS . "$actual_body") <(jq -cS . "$expected_body"))" ]; then
+  if [[ -z "$JSON" ]] || [ "$(cmp <(echo "$JSON" | jq .) <(echo "$EXPECTED_JSON" | jq .))" ]; then
     echo "----"
     echo "FAIL:"
-    echo "  actual:   $(jq -cS . "$expected_body")"
-    echo "  expected: $(jq -cS . "$expected_body")"
+    echo "  actual:   $(echo "$JSON" | jq .)"
```
beaugrantham-va, line 106: Note that this changes the behavior - previously it was compact (`-c`) and now it is pretty printed. Not sure if this was intentional.

beaugrantham-va (approved):

oauth-proxy/tests/scripts/assertions.sh
lukemarkwordtlibertyits: No that was not intentional. Will revert that behavior.

PR #202, opened 2020-12-16
Api 3459 default category check

lukemarkwordtlibertyits:
Preliminary PR to removing default routes. Checks for default api_category in the app config. Will use default routes if no default api_category then it will use the default routes.

[API3459-Tests.docx](https://github.com/department-of-veterans-affairs/vets-saml-proxy/files/5705316/API3459-Tests.docx)

### DIFF
--- a/oauth-proxy/index.js
+++ b/oauth-proxy/index.js
@@ -132,11 +132,6 @@ function buildApp(
   const { well_known_base_path } = config;
   const redirect_uri = `${config.host}${well_known_base_path}${config.routes.app_routes.redirect}`;
 
-  /**
-   * @deprecated - To be removed following AuthZ Server reorganization
-   */
-  const metadataRewrite = buildMetadataRewriteTable(config);
-
   const app = express();
   const router = new express.Router();
   // Express needs to know it is being ran behind a trusted proxy. Setting 'trust proxy' to true does a few things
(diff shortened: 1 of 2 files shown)


oauth-proxy/index.js
```diff
@@ -168,103 +163,112 @@ function buildApp(
     preflightContinue: true,
   });
 
-  // @deprecated - To be removed following AuthZ Server reorganization
-  router.options("/.well-known/*", corsHandler);
```
lukemarkwordtlibertyits, line 172: Does anyone know if this middleware is necessary? It is not in the broken out endpoints, but my tests are passing and I can hit the metadata endpoint.

oauth-proxy/index.js
```diff
@@ -168,103 +163,112 @@ function buildApp(
...
-    proxyRequestToOkta(req, res, issuer.metadata.jwks_uri, "GET")
-  );
-
-  // @deprecated - To be removed following AuthZ Server reorganization
-  router.get(config.routes.app_routes.userinfo, (req, res) =>
-    proxyRequestToOkta(req, res, issuer.metadata.userinfo_endpoint, "GET")
-  );
+  if (
+    !config.routes.categories.find((category) => category.api_category === "") >
+    0
```
crolarlibertyva, line 168: You don't need the " > 0" for this to work.

crolarlibertyva (changes requested):

crolarlibertyva:
Works for me. Tested with default route in place and not.

crolarlibertyva (approved):

beaugrantham-va (approved):

PR #205, opened 2020-12-18
Api 4051 bad error response

bdgrider:


.gitignore
```diff
@@ -120,6 +120,7 @@ $RECYCLE.BIN/
 build/
 /test-report.xml
 /saml-proxy/public/core.css
+/saml-proxy/sp-metadata.xml
```
crolarlibertyva, line 123: I think this wasn't meant to be here.

crolarlibertyva (changes requested):

.gitignore
crolarlibertyva: This is fine to keep here.

oauth-proxy/oauthHandlers/tokenHandlerStrategyClasses/tokenHandlerClient.js
```diff
@@ -26,6 +26,20 @@ class TokenHandlerClient {
       tokens = await this.getTokenResponseStrategy.getTokenResponse();
     } catch (error) {
       rethrowIfRuntimeError(error);
+      if (
+        error.error != undefined &&
+        error.error
+          .toString()
+          .includes("expected 200 OK, got: 401 Unauthorized")
```
beaugrantham-va, line 33: Interesting. So it's the OIDC client that returns this error. Understandable I guess.

This is the right approach (catch and rethrow an OAuth error) but I worry about using `error.error` as the value to compare here. It looks pretty free-form and subject to change (casing, punctuation, expected value(s)) at the discretion of the library owner.

It looks like value has a status code in it that might be safer.

```
{"error":"expected 200 OK, got: 401 Unauthorized","statusCode":401}
```

I recommend comparing `statusCode == 401` and then re-throwing a 401/invalid_client with a generic description (something like Client authentication failed.).

oauth-proxy/tests/scripts/token_tests.sh
```diff
@@ -208,7 +208,7 @@ do_token "$(jq \
 "$DIR"/assertions.sh --expect-status --status="$(cat "$curl_status")" --expected-status=401
 track_result
 
-# expect_json_body '{"error":"invalid_client", "error_description": "Invalid value for client_id parameter."}'
+expect_json_body '{"error":"invalid_client", "error_description": "Invalid value for client_id parameter."}'
```
beaugrantham-va, line 211: This test doesn't look like it should pass. It's expecting `Invalid value for client_id parameter.` but the actual error message is `Invalid value for client_id parameter..`? Can you double check that `expect_json_body` is working correctly?

beaugrantham-va (changes requested):

oauth-proxy/tests/scripts/token_tests.sh
```diff
@@ -208,7 +208,7 @@ do_token "$(jq \
...
```
beaugrantham-va, line 212: There is a second commented token test in this file that should be enabled.

oauth-proxy/tests/scripts/regression_tests.sh
```diff
@@ -48,7 +48,7 @@ fi
 if [ -z "$USER_EMAIL" ];
 then
   echo "ERROR - USER_EMAIL is a required parameter."
-  exit 1
+  exit 1mktemp
```
beaugrantham-va, line 51: Revert this change.

beaugrantham-va (changes requested):

crolarlibertyva (approved):

dbrown-va (approved):

beaugrantham-va (approved):

PR #207, opened 2020-12-22
Update project metadata

beaugrantham-va:
Update author and URLs.


dbrown-va:
Should we go ahead and tackle the SAML proxy side as well?
https://github.com/department-of-veterans-affairs/vets-saml-proxy/blob/master/saml-proxy/package.json#L6

dbrown-va (approved):

PR #209, opened 2020-12-22
API-3494 Add a request-id to log entries for better correlation

bdgrider:
Adding a request id to all requests using the cls-rtracer library, like SAML proxy.


beaugrantham-va (changes requested):
- [x] Update title
- [ ] Add description
- [ ] Link Jira story
- [ ] Resolve conflicts

oauth-proxy/package-lock.json
```diff
@@ -3,6 +3,9328 @@
   "version": "1.0.0",
   "lockfileVersion": 1,
   "requires": true,
+<<<<<<< HEAD
```
beaugrantham-va, line 6: This merge is incomplete.

oauth-proxy/tests/e2e.test.js
```diff
@@ -17,6 +17,7 @@ const {
 } = buildBackgroundServerModule("oauth-proxy test app");
 const { buildApp } = require("../index");
 const { encodeBasicAuthHeader } = require("../utils");
+require("cls-rtracer");
```
beaugrantham-va, line 20: This seems to be unnecessary.

beaugrantham-va (changes requested):

oauth-proxy/package-lock.json
bdgrider: Strange, I do not see this line in my dev environment and my branch says it is up to date.

beaugrantham-va: Strange. I pushed a clean merged copy.

oauth-proxy/logger.js
```diff
@@ -1,4 +1,5 @@
 const { createLogger, format, transports } = require("winston");
```
lukemarkwordtlibertyits, line 1: There are two loggers that concern us, morgan and winston. Morgan logs the requests, whereas winston is the logger that logs messages such as log.inf("yada yada"). For the request_id to be useful we need to add it to both the morgan and winston loggers. This PR only adds the request_id to the morgan logger, we need to add the winston logger as well.

The `winstonMiddleware` function in the saml-proxy/src/logger.ts module will be a helpful reference.

lukemarkwordtlibertyits (changes requested):

oauth-proxy/logger.js
lukemarkwordtlibertyits: Current winston logs look similar to below:
```
{
	"service": "oauth-proxy",
	"error": "invalid_grant",
	"message": "Could not refresh the client session with the provided refresh token invalid_grant (The refresh token is invalid or expired.)",
	"name": "OPError",
	"error_description": "The refresh token is invalid or expired.",
	"level": "error",
	"time": "2020-12-23T23:50:54.273Z"
}
```

lukemarkwordtlibertyits (approved):
Looks good

crolarlibertyva (approved):

beaugrantham-va:
My only nitpick is the underscore being inconsistent with the other properties that all use dashes.

```
oauth-proxy_1         |   "remote-address": "::ffff:127.0.0.1",
oauth-proxy_1         |   "status-code": "200",
oauth-proxy_1         |   "content-length": "2464",
oauth-proxy_1         |   "request_id": "72403c60-4e98-11eb-8763-930cdbfbd945"
```

Given that saml-proxy has the same behavior, I'm okay with it for consistency, but did want to call it out.

beaugrantham-va (approved):

PR #210, opened 2020-12-29
Api 3862 friendly error handling

lukemarkwordtlibertyits:
Created more descriptive saml proxy error handling. All use VA formatting.

# Changes

The following is a description of the various Saml Proxy error templates.

## Internal Failure

These errors consists of failures within the VA system. They may be from the Saml Proxy or another system such as MVI.

### Contents

- VA Formatting
- VA Tools not working message
- Request ID
 
internalFailure

## Error

These are general errors. It is usually caused by invalid input. They display messages to help the user trouble shoot their issue.

### Contents

- VA Formatting
- Error Description
- Request ID

error

## Sensitive Error

These errors contain information that should not be exposed publicaly, such as information about an attack being prevented.

### Contents

- VA Formatting
- General Error Message
- Request ID

sensitiveError

## ICN

This error occurs when the VA cannot retrieve an ICN from its internal IDPs.

### Contents

- VA Formatting
- Matching error message
- Request ID

matchingError

## Errors -> Templates

The following is a list of common errors and they template type that they are expected to render.

**Matching Error**

- ICN
- Internal Failure

**Invalid XML**

- Sensitive Error

**Modify SAML**

- Sensitive Error

**XSW Attacks**

- Sensitive Error

**Replay**

- Sensitive Error

**Empty Request**

- Sensitive Error

**Invalid Request**

- Sensitive Error

**404**

- Error


crolarlibertyva (approved):

beaugrantham-va:
Cool. This is slick.

Because all of these errors are going to end-users (vs developers) I think we can further group the mapping of the templates. Presenting an end-user with application specific errors like "invalid xml" or "empty relay state" in the description doesn't add any value, and potentially causes more confusion, given that there is nothing the user can do to fix it.

**500**
* Internal failure

**Matching Error**
* ICN

**404**
* Route not found

**Anything else**
* Your request could not be processed

beaugrantham-va:
Do we lose any error logging or data used for metrics (like HTTP status code) with this change?

beaugrantham-va:
We need to run the contact/help info ("What you can do") by someone. Those contacts or phone numbers might not be applicable for 400/404 errors.

lukemarkwordtlibertyits:
> Do we lose any error logging or data used for metrics (like HTTP status code) with this change?

That is a good question. Let me check.

beaugrantham-va (changes requested):
For internalFailure, error, and sensitiveError, let's direct users to submit a support ticket (https://developer.va.gov/support/contact-us) instead of calling the phone number.

esubocz-va:
> For internalFailure, error, and sensitiveError, let's direct users to submit a support ticket (https://developer.va.gov/support/contact-us) instead of calling the phone number.

would it be good to do a "submit support ticket (url>... _or_ call <phone-number>? or are we fully pulling away from that?

beaugrantham-va:
> > For internalFailure, error, and sensitiveError, let's direct users to submit a support ticket (https://developer.va.gov/support/contact-us) instead of calling the phone number.
> 
> would it be good to do a "submit support ticket `<url>`... _or_ call `<phone-number>` or are we fully pulling away from that?

The phone number that we currently advertise would not be able to provide support application errors. To the best of my knowledge it is only applicable for the one error case where ICN matching fails.

esubocz-va:
> > > For internalFailure, error, and sensitiveError, let's direct users to submit a support ticket (https://developer.va.gov/support/contact-us) instead of calling the phone number.
> > 
> > 
> > would it be good to do a "submit support ticket `<url>`... _or_ call `<phone-number>` or are we fully pulling away from that?
> 
> The phone number that we currently advertise would not be able to provide support for application errors. To the best of my knowledge it is only applicable for the one error case where ICN matching fails.

ahh okay, cool cool cool then 👍

bdgrider (commented):
All tests passed that I ran. Unsure of how to run a comment truncation attack that was mentioned in the testing doc (Test #10, which is the second occurrence of "Test Nine" in the testing document). A little more detail would be helpful on that one.

dbrown-va (approved):

PR #212, opened 2020-12-31
Updated regression tests 

bdgrider:
Updated the regression tests to revoke grants and then consent on the auth flow using changes made to the lighthouse-auth-utils bot.


oauth-proxy/tests/scripts/regression_tests.sh
```diff
@@ -147,6 +149,11 @@ assign_code() {
 # Pulling latest lighthouse-auth-utils docker image if necessary
 docker pull vasdvp/lighthouse-auth-utils:latest
 
+"$DIR"/okta_grants_tests.sh --host="$HOST"
```
beaugrantham-va, line 152: Is this a test or a setup for a test? If it's a test, it belongs in `# Start Tests` section.

beaugrantham-va (changes requested):
This branch has conflicts

oauth-proxy/tests/scripts/regression_tests.sh
bdgrider: I believe this needs to come before the other tests and the assign_code function in order to trigger the okta consenting pages. I moved the # Start Tests section comment above the okta_grants_tests.sh to make it clear this is the first test.

beaugrantham-va (approved):

PR #211, opened 2020-12-30
Api 3592 static token

crolarlibertyva:
Enables the use of static tokens to be retrieved during the refresh flow. 
To test the new functionality
-  pre-populate with 
`aws dynamodb put-item \
--table-name StaticTokens  \
--endpoint-url http://localhost:8000 \
--item \
    "{\"static_access_token\": {\"S\": \"$ACCESS_TOKEN\"}, \"static_refresh_token\": {\"S\": \"$REFRESH_TOKEN\"}, \"static_scopes\": {\"S\": \"$SCOPES\"}, \"static_expires_in\": {\"N\": \"3600\"}, \"static_icn\": {\"S\": \"$ICN\"}}"`
- exercise the refresh flow using the refresh token from above.

Note that in order for this functionality to be available an MR that creates a new dynamo table  has to be serviced. If the table is not available, the code will retain the old behavior.

PR #214, opened 2021-01-08
Api 3592 alternate cache static tokens implementation

dbrown-va:
https://vajira.max.gov/browse/API-3592

Enables the use of static tokens to be retrieved during the refresh flow.

Alternate approach to [211](https://github.com/department-of-veterans-affairs/vets-saml-proxy/pull/211) which checks DynamoDB with each token call.

A static refresh token of `987654321` is prestored for local development by the dynamo-migrations script.

Devops ticket created for the DynamoDB table creation.
department-of-veterans-affairs/lighthouse-devops-support#151


crolarlibertyva (approved):

bdgrider (approved):
Verified unit and regression tests pass, verified new dynamodb table logs that the table was created on start up, reviewed code.

beaugrantham-va (approved):

PR #216, opened 2021-01-15
API 4280 Verify redirect uri is present earlier.

bdgrider:
Added a simple check for redirect uri with an oauth2 compliant error message at the beginning of /authorization middleware function.

### DIFF
--- a/oauth-proxy/oauthHandlers/authorizeHandler.js
+++ b/oauth-proxy/oauthHandlers/authorizeHandler.js
@@ -13,9 +13,9 @@ const authorizeHandler = async (
   res,
   next
 ) => {
+  loginBegin.inc();
   const { state, client_id, aud, redirect_uri: client_redirect } = req.query;
 
-  loginBegin.inc();
   try {
     await checkParameters(
       state,
--- a/oauth-proxy/oauthHandlers/authorizeHandler.js
+++ b/oauth-proxy/oauthHandlers/authorizeHandler.js
@@ -14,18 +14,17 @@ const authorizeHandler = async (
   next
 ) => {
   const { state, client_id, aud, redirect_uri: client_redirect } = req.query;
-  if (client_redirect === undefined || client_redirect === "") {
-    res.status(400).json({
-      error: "invalid_client",
-      error_description:
-        "There was no redirect URI specified by the application.",
-    });
-    return next();
-  }
 
   loginBegin.inc();
   try {
-    await checkParameters(state, aud, issuer, logger, oktaClient);
+    await checkParameters(
+      state,
+      aud,
+      issuer,
+      logger,
+      oktaClient,
+      client_redirect
+    );
   } catch (err) {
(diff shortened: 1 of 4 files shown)


oauth-proxy/oauthHandlers/authorizeHandler.js
```diff
@@ -13,9 +13,17 @@ const authorizeHandler = async (
   res,
   next
 ) => {
-  loginBegin.inc();
   const { state, client_id, aud, redirect_uri: client_redirect } = req.query;
+  if (client_redirect === "") {
```
lukemarkwordtlibertyits, line 17: This will not catch undefined redirect_uris. ex: `curl -s \
  -w "%{http_code}" \
  -o "$curl_body" \
  "$HOST/authorization?client_id=$CLIENT_ID&scope=$SCOPE&response_type=code&aud=default" `

lukemarkwordtlibertyits (changes requested):

oauth-proxy/oauthHandlers/authorizeHandler.js
```diff
@@ -13,9 +13,17 @@ const authorizeHandler = async (
...
+  if (client_redirect === undefined || client_redirect === "") {
+    res.status(400).json({
+      error: "invalid_client",
+      error_description:
+        "There was no redirect URI specified by the application.",
+    });
+    return next();
+  }
 
+  loginBegin.inc();
```
beaugrantham-va, line 26: Move this back up to the top.

It's an important metric to capture the count of _all_ login attempts, regardless of why they may not complete.

beaugrantham-va (dismissed):

oauth-proxy/oauthHandlers/authorizeHandler.js
```diff
@@ -13,9 +13,17 @@ const authorizeHandler = async (
...
   try {
     await checkParameters(state, aud, issuer, logger, oktaClient);
```
dbrown-va, line 28: We are doing a number of other client supplied parameter checks here.

Moving the `client_redirect` check could be worth while.  That would also address the `loginBegin.inc()` point Beau mentions above.

oauth-proxy/oauthHandlers/authorizeHandler.js
```diff
@@ -13,9 +13,17 @@ const authorizeHandler = async (
   res,
   next
 ) => {
-  loginBegin.inc();
   const { state, client_id, aud, redirect_uri: client_redirect } = req.query;
+  if (client_redirect === undefined || client_redirect === "") {
```
lukemarkwordtlibertyits, line 17: A truthy check may be better for this statement

```
if(!client_redirect)
```

dbrown-va (approved):

lukemarkwordtlibertyits (approved):
looks good

crolarlibertyva (approved):

PR #219, opened 2021-01-19
Api 3839 refactor dynamo client

crolarlibertyva:
Dynamo client is now a class that only uses the document api for dynamo. Use of the client changed accordingly.


oauth-proxy/dynamo_client.js
```diff
@@ -1,222 +1,133 @@
...
-  };
-
-  return new Promise((resolve, reject) => {
-    client.query(params, (err, data) => {
-      if (err) {
-        reject(err);
-      } else {
-        resolve(data.Items[0]);
-      }
+  savePayloadToDynamo(payload, tableName) {
```
beaugrantham-va, line 23: Please add docs to all of these.

beaugrantham-va (changes requested):
In general it's not clear to me when I would use `getPayloadFromDynamo` vs `queryFromDynamo`

lukemarkwordtlibertyits:
Overall I do not have any suggestions to this branch. I like the refactoring. Having a DynamoClient that was dependent on a Dynamo Object was a strange pattern. I am glad this takes care of it. Also, the old DynamoClient naming conventions were obscured as we added more tables. For example, `saveToDynamo` saved to the refresh token table where as `saveToDynamoAccessToken` saved to the client credentials table. Its good that the methods are more generic now. Finally, this PR increased the overall code coverage from 84% to 87%!

oauth-proxy/README.md
```diff
@@ -30,6 +30,12 @@ Start the Oauth Proxy
 npm start
 ```
 
+Generate jsdocs for Oauth Proxy
+```
+npm run-script jsdoc
```
lukemarkwordtlibertyits, line 35: should the out folder be in gitignore?

oauth-proxy/oauthHandlers/redirectHandler.js
```diff
@@ -39,19 +36,38 @@ const redirectHandler = async (
...
+      {
+        state: state,
+      },
       config.dynamo_table_name
     );
-
-    const params = new URLSearchParams(req.query);
-    loginEnd.inc();
-    res.redirect(`${document.redirect_uri.S}?${params.toString()}`);
+    if (document && document.Item && document.Item.redirect_uri) {
```
lukemarkwordtlibertyits, line 45: could be 
```
if(document?.Item?.redirect_uri)
```

I guess it depends on preference

oauth-proxy/oauthHandlers/redirectHandler.js
```diff
@@ -39,19 +36,38 @@ const redirectHandler = async (
...
     logger.error("Failed to redirect to the OAuth client application", error);
-    return next(error); // This error is unrecoverable because we can't look up the original redirect.
+    // This error is unrecoverable because we can't look up the original redirect.
+    res.status(400).json({
+      error: "invalid_request",
+      error_description:
+        "Unable to look up the original redirect for the given state",
+    });
   }
+  // Only would reach here upon an error
```
lukemarkwordtlibertyits, line 69: I'm not sure this is true. res.status will not return a function, with that being said, this next statement will have no outwards issues associated with it. You will probably see one of these errors in the logs

```
(node:6) UnhandledPromiseRejectionWarning: TypeError: Cannot read property 'headers' of undefined
oauth-proxy_1         |     at setProxyResponse (/home/node/index.js:96:18)
oauth-proxy_1         |     at /home/node/index.js:128:9
oauth-proxy_1         |     at processTicksAndRejections (internal/process/task_queues.js:97:5)
oauth-proxy_1         | (node:6) UnhandledPromiseRejectionWarning: Unhandled promise rejection. This error originated either by throwing inside of an async function without a catch block, or by rejecting a promise which was not handled with .catch(). To terminate the node process on unhandled promise rejection, use the CLI flag `--unhandled-rejections=strict` (see https://nodejs.org/api/cli.html#cli_unhandled_rejections_mode). (rejection id: 7)
oauth-proxy_1         | (node:6) [DEP0018] DeprecationWarning: Unhandled promise rejections are deprecated. In the future, promise rejections that are not handled will terminate the Node.js process with a non-zero exit code.
```

oauth-proxy/README.md
crolarlibertyva: yes.

oauth-proxy/oauthHandlers/redirectHandler.js
crolarlibertyva: I like that better.

oauth-proxy/oauthHandlers/redirectHandler.js
crolarlibertyva: Will check it out. This inadvertently fixed the 500 error that was referred to during grooming.

oauth-proxy/oauthHandlers/redirectHandler.js
crolarlibertyva: Had to revert to get it to run in docker.

oauth-proxy/oauthHandlers/redirectHandler.js
crolarlibertyva: Didn't see this in logs for regression test.

oauth-proxy/oauthHandlers/redirectHandler.js
crolarlibertyva: We should consider an upgraded docker base image.

oauth-proxy/dynamo_client.js
```diff
@@ -1,222 +1,172 @@
...
-  };
-
-  return new Promise((resolve, reject) => {
-    dynamoDb.dbDocClient.scan(params, function (err, data) {
-      if (err) {
-        reject(err);
-      } else {
-        resolve(data);
-      }
+    params.UpdateExpression = params.UpdateExpression.slice(0, -1);
```
beaugrantham-va, line 158: Should this be updated like the other to remove the arbitrary slice?

oauth-proxy/oauthHandlers/redirectHandler.js
```diff
@@ -39,19 +36,38 @@ const redirectHandler = async (
...
+      loginEnd.inc();
+      res.redirect(`${document.redirect_uri}?${params.toString()}`);
+    } else {
+      logger.error(
+        "Failed to get the redirect for the OAuth client application"
+      );
+      res.status(400).json({
+        error: "invalid_request",
+        error_description:
+          "Unable to look up the original redirect for the given state",
```
beaugrantham-va, line 57: I would opt to be less verbose here (and below). Something like "Invalid state". I feel like this message exposes application detail/logic that is unnecessary to the user.

lukemarkwordtlibertyits (approved):
Overall looks good. This refactor helps clean the repository quite a bit.

dbrown-va (approved):
Confirmed running locally and the earlier changes requested have been implemented.

PR #221, opened 2021-01-26
Api 4185 503 error handling for validation endpoint

lukemarkwordtlibertyits:
# API4185

Changed the Oauth-Proxy to handle validation errors as 503 to indicate the service is temporarily unavailable.

## Should a scripted integration test be changed or created for this use case?

The following is a list of the current Token Handler Tests:
    - Token Handler code happy path
    - Token Handler expired code
    - Token Handler invalid code
    - Token Handler code path invalid client id
    - Revoke active token happy path
    - Token Handler refresh happy path
    - Token Handler invalid refresh Token
    - Token Handler refresh path invalid client id
    - Client Credentials happy path
    - Token Handler invalid strategy
    - Token Handler missing grant_type

There is currently no test for Validation Errors, because it is not possible to force Vets-API to throw an error on validation. Work arounds would be to mock this out or to point the validation post endpoint in the Oauth Proxy config to a bad url. Although this not possible to do with the scripted integration tests, I did add a e2e test to reflect this change, which mocks the validation endpoint.

It may be worth investigating how to handle this usecase in the integration tests.

## How to Test

Set `"validate_post_endpoint": "bad"` in the oauth proxy config.

#

 API4185 Tests

Vets-API Validation errors should be 503 not 500.

## Happy Path

Vets-API Validation successes should not change.

### Test Steps

1. Go through the auth flow with a functional Vets-API reference and the launch/patient scope set. This can be done by running the regression tests script.

### Expected Results

- All regression tests should Pass
    - Confirmed

## System Error should be 503

500 errors should be changed to 503.

### Test Steps

1. Point the veteran validation endpoint to a bad instance.
2. Go through the auth flow.

### Expected Results

- 503 error should be returned.
- Retry-After header should be included with a value of 300.

Both Passed

```
HTTP/1.1 503 Service Unavailable
X-Powered-By: Express
Retry-After: 300
Content-Type: application/json; charset=utf-8
Content-Length: 31
ETag: W/"1f-T6srg3j0SRbz7gfrP4Xi+ouCtAM"
Date: Tue, 26 Jan 2021 22:36:04 GMT
Connection: keep-alive
Keep-Alive: timeout=75

{"error":"Service_Unavailable"}
```

### DIFF
--- a/oauth-proxy/index.js
+++ b/oauth-proxy/index.js
@@ -228,12 +228,11 @@ function buildApp(
         error_description: "Invalid or unsupported content-type",
       });
     } else if (err.statusCode && err.statusCode === 503) {
-      res
+      res.set("Retry-After", "300")
         .status(503)
         .json({
-          error: "server_error",
-        })
-        .headers("Retry-After", "300");
+          error: "Service_Unavailable",
+        });
     } else if (error && error_description) {
       res.status(500).json({
         error: "server_error",
--- a/oauth-proxy/index.js
+++ b/oauth-proxy/index.js
@@ -229,7 +229,7 @@ function buildApp(
       });
     } else if (err.statusCode && err.statusCode === 503) {
       res.set("Retry-After", "300").status(503).json({
-        error: "Service_Unavailable",
+        error: "temporarily_unavailable",
       });
     } else if (error && error_description) {
       res.status(500).json({
--- a/oauth-proxy/index.js
+++ b/oauth-proxy/index.js
@@ -228,11 +228,9 @@ function buildApp(
         error_description: "Invalid or unsupported content-type",
       });
     } else if (err.statusCode && err.statusCode === 503) {
-      res.set("Retry-After", "300")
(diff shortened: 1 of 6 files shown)


oauth-proxy/index.js
```diff
@@ -227,6 +227,10 @@ function buildApp(
         error: "invalid_request",
         error_description: "Invalid or unsupported content-type",
       });
+    } else if (err.statusCode && err.statusCode === 503) {
+      res.set("Retry-After", "300").status(503).json({
+        error: "Service_Unavailable",
```
beaugrantham-va, line 232: This needs to be a value defined in the RFC.

It looks like the two values that fit are `server_error` or `temporarily_unavailable`.

beaugrantham-va (changes requested):

crolarlibertyva (commented):
Confirmed 503 status.

beaugrantham-va (approved):

PR #222, opened 2021-01-27
Removed Corner Case e2e test

lukemarkwordtlibertyits:


dbrown-va (approved):
Confirmed removal still passes coverage threshold.

PR #223, opened 2021-01-28
Api 4758 auth0 wsfed saml2

crolarlibertyva:
Replaces dependency from lighthouse-passport-wsfed-saml2 to the original upstream of auth0/passport-wsfed-saml2.


crolarlibertyva:
Fixing tests will take longer than expected.

PR #225, opened 2021-02-03
Api 4620 missing relay

crolarlibertyva:


lukemarkwordtlibertyits (dismissed):
Confirmed that this does not break functionality. Cannot test MHV without VPN. Should be safe to merge and then test in QA.

dbrown-va (approved):
Looks good with the addition of unit tests

lukemarkwordtlibertyits (approved):
Confirmed previous functionality is working

PR #226, opened 2021-02-05
API-4749 SMART launch context with authorization code flow

dbrown-va:
Adds SMART launch context storing capability for the authorization code flow for configured IDP's.

https://vajira.max.gov/browse/API-4749

Some background on SMART launch context below
http://www.hl7.org/fhir/smart-app-launch/scopes-and-launch-context/#apps-that-launch-from-the-ehr

Main changes to look for:
- Launch context should only be stored if the base `launch` scope is provided when using the authorization code flow
- Client Credentials flow should continue unchanged
- Refresh token calls should check if a launch context is associated and if so create an entry in the Launch Context table

### DIFF
--- a/oauth-proxy/cli.js
+++ b/oauth-proxy/cli.js
@@ -79,6 +79,12 @@ function processArgs() {
         description:
           "Okta IDP identifier to be added as a query param (idp) if not specified by user in authorize request",
       },
+      smart_launch_idps: {
+        description:
+          "An comma separated list of idp keys that allow storing of launch context",
+        required: true,
+        default: "",
+      },
       sentry_dsn: {
         description: "URL of the sentry project to send errors",
         required: false,
--- a/oauth-proxy/cli.js
+++ b/oauth-proxy/cli.js
@@ -79,12 +79,6 @@ function processArgs() {
         description:
           "Okta IDP identifier to be added as a query param (idp) if not specified by user in authorize request",
       },
-      smart_launch_idps: {
-        description:
-          "A comma separated list of idp keys that allow storing of launch context",
-        required: true,
-        default: "",
-      },
       sentry_dsn: {
         description: "URL of the sentry project to send errors",
         required: false,
--- a/oauth-proxy/cli.js
+++ b/oauth-proxy/cli.js
@@ -81,7 +81,7 @@ function processArgs() {
       },
       smart_launch_idps: {
         description:
-          "An comma separated list of idp keys that allow storing of launch context",
+          "A comma separated list of idp keys that allow storing of launch context",
         required: true,
         default: "",
       },
--- a/oauth-proxy/oauthHandlers/authorizeHandler.js
+++ b/oauth-proxy/oauthHandlers/authorizeHandler.js
@@ -150,7 +150,7 @@ const checkParameters = async (
   }
 
   if (
-    scopes !== undefined &&
+    scopes &&
     scopes.split(" ").filter((scope) => scope.includes("launch/")).length > 1
   ) {
     logger.error("Multiple launch contexts were specified.");
--- a/oauth-proxy/oauthHandlers/authorizeHandler.js
+++ b/oauth-proxy/oauthHandlers/authorizeHandler.js
@@ -22,8 +22,7 @@ const authorizeHandler = async (
       issuer,
       logger,
       oktaClient,
-      client_redirect,
-      req.query.scope
+      client_redirect
     );
   } catch (err) {
     if (err.status == 500) {
@@ -103,8 +102,7 @@ const checkParameters = async (
   issuer,
   logger,
   oktaClient,
-  client_redirect,
-  scopes
+  client_redirect
 ) => {
   if (!client_redirect) {
     logger.error("No valid redirect_uri was found.");
@@ -147,18 +145,6 @@ const checkParameters = async (
         expected: serverAudiences,
       });
     }
-  }
-
-  if (
-    scopes &&
-    scopes.split(" ").filter((scope) => scope.includes("launch/")).length > 1
-  ) {
-    logger.error("Multiple launch contexts were specified.");
-    throw {
-      status: 400,
-      error: "invalid_request",
-      error_description: "Only 1 launch context can be specified.",
-    };
   }
 };
 
--- a/oauth-proxy/oauthHandlers/authorizeHandler.js
+++ b/oauth-proxy/oauthHandlers/authorizeHandler.js
@@ -22,7 +22,8 @@ const authorizeHandler = async (
       issuer,
       logger,
       oktaClient,
-      client_redirect
+      client_redirect,
+      req.query.scope
     );
   } catch (err) {
     if (err.status == 500) {
@@ -104,7 +105,8 @@ const checkParameters = async (
   issuer,
   logger,
   oktaClient,
-  client_redirect
+  client_redirect,
+  scopes
 ) => {
   if (!client_redirect) {
     logger.error("No valid redirect_uri was found.");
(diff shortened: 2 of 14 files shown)


oauth-proxy/cli.js
```diff
@@ -79,6 +79,12 @@ function processArgs() {
         description:
           "Okta IDP identifier to be added as a query param (idp) if not specified by user in authorize request",
       },
+      smart_launch_idps: {
+        description:
+          "An comma separated list of idp keys that allow storing of launch context",
```
beaugrantham-va, line 84: `s/An /A`

oauth-proxy/oauthHandlers/tokenHandlerStrategyClasses/saveDocumentStrategies/saveDocumentStateStrategy.js
```diff
@@ -27,6 +28,30 @@ class SaveDocumentStateStrategy {
...
+          expires_on: decodedToken.exp,
+        };
+
+        await this.dynamoClient.savePayloadToDynamo(
+          payload,
+          this.config.dynamo_launch_context_table
+        );
+      }
+    } catch (error) {
+      this.logger.error("Could not save the launch context in DynamoDB", error);
```
beaugrantham-va, line 53: I think this should be a 500 scenario.

If a token is intended for a specific launch context then we can't lose that.

oauth-proxy/oauthHandlers/tokenHandlerStrategyClasses/tokenHandlerClient.js
```diff
@@ -55,23 +55,30 @@ class TokenHandlerClient {
...
 
     let state;
+    let launch;
     if (document && tokens) {
       await this.saveDocumentToDynamoStrategy.saveDocumentToDynamo(
         document,
         tokens
       );
       state = document.state || null;
+      launch = document.launch || null;
```
beaugrantham-va, line 65: I think it's safe to drop the `|| null` here for launch. It's there explicitly for `state` to ensure that it gets populated in the token response whether or not there is a value.

oauth-proxy/oauthHandlers/tokenHandlerStrategyClasses/tokenHandlerClient.js
```diff
@@ -55,23 +55,30 @@ class TokenHandlerClient {
...
     //Creates a Token Response
     const tokenResponseBase = translateTokenSet(tokens);
     let decoded = jwtDecode(tokens.access_token);
     if (decoded.scp != null && decoded.scp.indexOf("launch/patient") > -1) {
-      let patient = await this.getPatientInfoStrategy.createPatientInfo(
-        tokens,
-        decoded
-      );
+      let patient;
+      if (launch) {
```
beaugrantham-va, line 74: This will not match if `launch == 0`. In combination with dropping the `|| null` this should probably change to `launch !== undefined`.

oauth-proxy/cli.js
dbrown-va: Updated!

oauth-proxy/oauthHandlers/tokenHandlerStrategyClasses/tokenHandlerClient.js
dbrown-va: Updated to remove `|| null`

oauth-proxy/oauthHandlers/tokenHandlerStrategyClasses/tokenHandlerClient.js
dbrown-va: Updated to `launch !== undefined` as recommended.

dbrown-va:
Added Devops PRs:

- department-of-veterans-affairs/devops#8484
- department-of-veterans-affairs/lighthouse-oauth-proxy-configs#14

oauth-proxy/oauthHandlers/tokenHandlerStrategyClasses/saveDocumentStrategies/saveDocumentStateStrategy.js
dbrown-va: Agreed.  Updated to throw a 500 error in addition to log..

oauth-proxy/oauthHandlers/authorizeHandler.js
```diff
@@ -137,6 +150,18 @@ const checkParameters = async (
       });
     }
   }
+
+  if (
+    scopes !== undefined &&
```
lukemarkwordtlibertyits, line 155: truthy checks are best here. ie 
`if(scopes && ...)`

oauth-proxy/oauthHandlers/tokenHandlerStrategyClasses/tokenHandlerClient.js
```diff
@@ -55,29 +55,36 @@ class TokenHandlerClient {
...
-        statusCode: 200,
-        responseBody: { ...tokenResponseBase, patient, state },
-      };
+    let responseBody = { ...tokenResponseBase, state };
+    if (decoded.scp != null) {
+      let filteredScopes = decoded.scp.filter((scp) => scp.includes("launch/"));
+      if (filteredScopes.length > 0) {
+        let launchField = filteredScopes[0].split("/")[1];
+        //Needed for patient ICN lookup
+        if (launchField == "patient" && launch === undefined) {
```
lukemarkwordtlibertyits, line 78: Also would consider a truthy check here
`if (launchField == "patient" && !launch)`

lukemarkwordtlibertyits (dismissed):
Have confirmed functionality locally. The test coverage has increased slightly. I would consider making the truthy check changes. Although the undefined checks work, I think truthy checks may be best practice. Also it is less code.

oauth-proxy/oauthHandlers/authorizeHandler.js
dbrown-va: Switched to truthy check as recommended.
https://github.com/department-of-veterans-affairs/vets-saml-proxy/pull/226/commits/15f5cdf6dd6ab7a39beb1fcb3648875fa25702ff

oauth-proxy/oauthHandlers/tokenHandlerStrategyClasses/tokenHandlerClient.js
dbrown-va: With the updates from further EHR Launch context discovery, this block has been updated to no longer have the `launch` undefined check.

oauth-proxy/oauthHandlers/authorizeHandler.js
```diff
@@ -137,6 +148,18 @@ const checkParameters = async (
       });
     }
   }
+
+  if (
+    scopes &&
+    scopes.split(" ").filter((scope) => scope.includes("launch/")).length > 1
```
beaugrantham-va, line 154: I don't know if this is still valid, after our review of the spec and sample apps.

It doesn't cause harm since we only support a single `launch/*` scope but may bite us later if/when we do support more.

I'd be in favor of removing this until we get to that point.

oauth-proxy/oauthHandlers/tokenHandlerStrategyClasses/saveDocumentStrategies/saveDocumentStateStrategy.js
```diff
@@ -27,6 +28,34 @@ class SaveDocumentStateStrategy {
         error
       );
     }
+
+    try {
+      if (document.launch && tokens.access_token) {
```
beaugrantham-va, line 33: I recommend also checking that the token contains the `launch` scope here.

oauth-proxy/oauthHandlers/authorizeHandler.js
dbrown-va: Agreed.  If it was valid in the spec, I guess maybe it would be another Base64 encoded `launch` with the assumption that it contains each of the `launch/*` scoped items?

I will remove at this time.

oauth-proxy/oauthHandlers/tokenHandlerStrategyClasses/saveDocumentStrategies/saveDocumentStateStrategy.js
dbrown-va: Agreed.  I've added launch scope checking as an additional check.

beaugrantham-va (dismissed):

beaugrantham-va (dismissed):

beaugrantham-va (approved):

PR #228, opened 2021-02-09
Api 4620 missing relay logging session

crolarlibertyva:
Adds some warnings around authnRequest not being available via the session.


lukemarkwordtlibertyits (approved):
Tests work. The code should not be an issue.

PR #232, opened 2021-02-11
Use docker image version 1.1.2 for regression test script in oauith-proxy

crolarlibertyva:
Use the last working docker image version for regression testing.


beaugrantham-va (approved):
Approved. Will still investigate the problem with 1.1.3.

PR #235, opened 2021-02-15
API-5200 - Add health/system authz route

beaugrantham-va:
Add new `health/system` route for consumers using client_credentials to access Health APIs.

This route will replace the single purpose `health-insurance` route. Lighthouse will work with existing consumers (1) to migrate them onto `health/system`.

Scopes will be assigned by policy/rule. This allows all system consumers to use the same authz route.


lukemarkwordtlibertyits (approved):
Have confirmed the CC flow works on this endpoint.

PR #236, opened 2021-02-15
Api 4962 local store client info

crolarlibertyva:
- Uses the client data-store, for validating /authorization requests. Reference https://github.com/department-of-veterans-affairs/vets-saml-proxy/blob/408455d3629d4118547df3083b7dbea638abd833/oauth-proxy/dynamo_schema.js#L145

To exercise the new functionality make the following changes to you local setup

- modify the contents of database with a valid client at https://github.com/department-of-veterans-affairs/vets-saml-proxy/blob/408455d3629d4118547df3083b7dbea638abd833/oauth-proxy/dynamo_schema.js#L240
- modify your local dev-config.js by altering one of the routes as follows. In this case the '/veteran-verification/v1' route will reference the local clients table.
`      {
        "api_category": "/veteran-verification/v1",
        "upstream_issuer": "https://deptva-eval.okta.com/oauth2/aus7y0sefudDrg2HI2p7",
        "manage_endpoint": "https://staging.va.gov/account",
        "client_store": "local"
      },
`

### DIFF
--- a/oauth-proxy/index.js
+++ b/oauth-proxy/index.js
@@ -18,6 +18,7 @@ const { jwtAuthorizationHandler } = require("./jwtAuthorizationHandler");
 const oauthHandlers = require("./oauthHandlers");
 const { configureTokenValidator } = require("./tokenValidation");
 const rTracer = require("cls-rtracer");
+const { SlugHelper } = require("./slug_helper");
 
 const openidMetadataWhitelist = [
   "issuer",
@@ -69,7 +70,6 @@ function filterProperty(object, property) {
 
 function buildApp(
   config,
-  oktaClient,
   dynamoClient,
   validateToken,
   isolatedIssuers,
@@ -91,6 +91,7 @@ function buildApp(
       },
     });
   }
+  const slugHelper = new SlugHelper(config);
 
   const setProxyResponse = (response, targetResponse) => {
     if (response.headers !== undefined) {
@@ -283,6 +284,7 @@ function buildApp(
           service_issuer,
           dynamoClient,
           okta_client,
+          slugHelper,
           req,
           res,
           next
@@ -353,12 +355,6 @@ function buildApp(
 }
 
 function startApp(config, isolatedIssuers) {
-  const oktaClient = new okta.Client({
-    orgUrl: config.okta_url,
-    token: config.okta_token,
-    requestExecutor: new okta.DefaultRequestExecutor(),
-  });
-
   const isolatedOktaClients = {};
   if (config.routes && config.routes.categories) {
     Object.entries(config.routes.categories).forEach(
@@ -388,7 +384,6 @@ function startApp(config, isolatedIssuers) {
   );
   const app = buildApp(
     config,
-    oktaClient,
     dynamoClient,
     validateToken,
     isolatedIssuers,
--- a/oauth-proxy/index.js
+++ b/oauth-proxy/index.js
@@ -178,6 +178,7 @@ function buildApp(
     const service_issuer = isolatedIssuers[isolatedOktaConfig.api_category];
     buildMetadataForOpenIdConfiguration(
       isolatedOktaConfig.api_category,
+      isolatedOktaConfig.manage_endpoint,
       app_routes,
       service_issuer,
       okta_client
@@ -246,6 +247,7 @@ function buildApp(
 
   function buildMetadataForOpenIdConfiguration(
     api_category,
+    manage_endpoint,
     app_routes,
     service_issuer,
     okta_client
@@ -310,9 +312,12 @@ function buildApp(
         .catch(next);
     });
 
-    router.get(api_category + app_routes.manage, (req, res) =>
-      res.redirect(config.manage_endpoint)
-    );
+    if (manage_endpoint) {
+      router.get(api_category + app_routes.manage, (req, res) =>
+        res.redirect(manage_endpoint)
+      );
+    }
+
     router.get(api_category + app_routes.jwks, (req, res) =>
       proxyRequestToOkta(req, res, service_issuer.metadata.jwks_uri, "GET")
     );
--- a/oauth-proxy/index.js
+++ b/oauth-proxy/index.js
@@ -179,7 +179,6 @@ function buildApp(
     buildMetadataForOpenIdConfiguration(
       app_category,
       app_routes,
-      isolatedOktaConfig.enable_consent_endpoint,
       service_issuer,
       okta_client
     );
@@ -248,7 +247,6 @@ function buildApp(
   function buildMetadataForOpenIdConfiguration(
     app_category,
     app_routes,
-    enable_consent_endpoint,
     service_issuer,
     okta_client
   ) {
@@ -353,7 +351,7 @@ function buildApp(
       );
     });
 
-    if (enable_consent_endpoint) {
+    if (app_category.enable_consent_endpoint) {
       router.delete(
         api_category + app_routes.grants,
         async (req, res, next) => {
--- a/oauth-proxy/index.js
+++ b/oauth-proxy/index.js
@@ -280,13 +280,16 @@ function buildApp(
     router.get(api_category + app_routes.authorize, async (req, res, next) => {
       await oauthHandlers
         .authorizeHandler(
-          config,
           redirect_uri,
           logger,
           service_issuer,
           dynamoClient,
           okta_client,
           slugHelper,
+          api_category,
+          config.dynamo_table_name,
+          config.dynamo_clients_table,
+          config.idp,
           req,
           res,
           next
--- a/oauth-proxy/index.js
+++ b/oauth-proxy/index.js
@@ -173,12 +173,11 @@ function buildApp(
   });
 
   const app_routes = config.routes.app_routes;
-  Object.entries(config.routes.categories).forEach(([, isolatedOktaConfig]) => {
-    const okta_client = isolatedOktaClients[isolatedOktaConfig.api_category];
-    const service_issuer = isolatedIssuers[isolatedOktaConfig.api_category];
+  Object.entries(config.routes.categories).forEach(([, app_category]) => {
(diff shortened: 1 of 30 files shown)


oauth-proxy/oauthHandlers/authorizeHandler.js
```diff
@@ -15,6 +15,18 @@ const authorizeHandler = async (
 ) => {
   loginBegin.inc();
   const { state, client_id, aud, redirect_uri: client_redirect } = req.query;
+  const app_category = config.routes.categories.find((item) => {
```
beaugrantham-va, line 18: What is this block of code (lines 18 - 29)?

crolarlibertyva: Uses the route to determine with app category, is used to determine if the route is configured for local lookup of clients.

oauth-proxy/oauthHandlers/authorizeHandler.js
```diff
@@ -151,4 +176,49 @@ const checkParameters = async (
...
+      config.dynamo_clients_table
+    );
+    if (clientInfo.Item) {
+      clientInfo = clientInfo.Item;
+    } else {
+      res.status(400).json({
+        error: "invalid_client",
+        error_description: "Invalid client for the authorization path",
+      });
+      return next();
```
beaugrantham-va, line 202: I'm not sure this pattern is correct... or please correct me if I am misunderstanding it.

Returning next() here is not the same as returning next() as a result of calling `localValidateClient` (https://github.com/department-of-veterans-affairs/vets-saml-proxy/pull/236/files#diff-63a76133f7b10311dfa0d9335a8fa16c23e5d16568c0d2500374041f92988c17R52). The calling code will continue to execute because there is nothing that interrupts it.

oauth-proxy/oauthHandlers/authorizeHandler.js
```diff
@@ -36,33 +48,46 @@ const authorizeHandler = async (
     return next();
   }
 
-  try {
-    const oktaApp = await oktaClient.getApplication(client_id);
-    if (
-      oktaApp.settings.oauthClient.redirect_uris.indexOf(client_redirect) === -1
-    ) {
+  if (app_category.client_store && app_category.client_store === "local") {
```
beaugrantham-va, line 51: `app_category.client_store` needs to be added to cli.js.

oauth-proxy/oauthHandlers/authorizeHandler.js
crolarlibertyva: Yes. Was fooled by the output on the screen. Even though the status got populated with the 404 the code didn't interrupt. Will have to change this.

lukemarkwordtlibertyits:
# API-4962 Code Review

## Confirm Happy Path is working.

- Make Authorization request.
- client_store = local
- client_store has redirect `localhost:8080/redirect`

### Expected results

- [x] Redirect to authorization server 

[ ] Confirm Client Redirect Miss is working.

- Make Authorization request.
- client_store = local
- client_store does not have redirect `localhost:8080/redirect`

expected results

- [x] expected json
```json
{
    error	"invalid_client",
    error_description	"The redirect URI specified by the application does not match any of the registered redirect URIs. Erroneous redirect URI: http://localhost:8080/redirect"
}
```

## Client miss in client_store

- Make Authorization request.
- client_store = local
- specified client not in client_store

expected results

- [ ] expected json
```json
{
    error: "invalid_client",
    error_description: "Invalid client for the authorization path",
}
```

## No Clients in client_store

- Make Authorization request.
- client_store = local
- no clients client_store

expected results

- [ ] expected json
```json
{
    error: "invalid_client",
    error_description: "The client specified by the application is not valid."
}
```

## No Okta Redirect URI

- Make Authorization request.
- client_store = local
- redirect uri not registered with okta

expected results

- [ ] expected json
```json
{
    error: "invalid_client",
    error_description: "The redirect URI specified by the application does not match any of the registered redirect URIs. Erroneous redirect URI: ${client_redirect}"
}
```

## No Okta Client Id

- Make Authorization request.
- client_store = local
- client id not registered with okta

expected results

- [ ] expected json
```json
{
    error: "invalid_client",
        error_description: "The client specified by the application is not valid.",
}
```

oauth-proxy/oauthHandlers/authorizeHandler.js
lukemarkwordtlibertyits, line 18: I wonder if this block of logic is necessary. We know what the api_category is when we instantiate this handler. Why couldn't we pass it into the function call?

I It would create an even longer method signature. We cant take the config object out as we use it in a couple other places, but this handler is getting pretty long, the tradeoff of having too many arguments for less logic in the method may be worth it. Thoughts?

oauth-proxy/oauthHandlers/authorizeHandler.js
```diff
@@ -36,33 +48,71 @@ const authorizeHandler = async (
...
```
lukemarkwordtlibertyits, line 51: This if statement could be replaced by a helper method.

```
try{
  checkRedirect
}catch(er) {
  res.status(err.status).json({
    error: err.error,
    error_description: err.error_description
  })  
  next()
}
```

I think it would make this method more readable.

lukemarkwordtlibertyits:
# API-4962 Code Review

## Confirm Happy Path is working.

- Make Authorization request.
- client_store = local
- client_store has redirect `localhost:8080/redirect`

### Expected results

- [x] Redirect to authorization server 

## Confirm Client Redirect Miss is working.

- Make Authorization request.
- client_store = local
- client_store does not have redirect `localhost:8080/redirect`

expected results

- [x] expected json
```json
{
    "error":	"invalid_client",
    "error_description": "The redirect URI specified by the application does not match any of the registered redirect URIs. Erroneous redirect URI: http://localhost:8080/redirect"
}
```

## Client miss in client_store

- Make Authorization request.
- client_store = local
- specified client not in client_store

expected results

- [x] expected json
```json
{
    "error": "invalid_client",
    "error_description": "Invalid client for the authorization path",
}
```

## No Clients in client_store

- Make Authorization request.
- client_store = local
- force dynamo error 

expected results

- [x] expected json
```json
{
    "error": "invalid_client",
    "error_description": "The client specified by the application is not valid."
}
```

## No Okta Redirect URI

- Make Authorization request.
- client_store = null
- redirect uri not registered with okta

expected results

- [x] expected json
```json
{
    "error": "invalid_client",
    "error_description": "The redirect URI specified by the application does not match any of the registered redirect URIs. Erroneous redirect URI: ${client_redirect}"
}
```

## No Okta Client Id

- Make Authorization request.
- client_store = null
- client id not registered with okta

expected results

- [x] expected json
```json
{
    "error": "invalid_client",
    "error_description": "The client specified by the application is not valid.",
}
```

I have confirmed expected functionality. My other comments were only suggestions and thoughts. I would be ok approving either way.

oauth-proxy/oauthHandlers/authorizeHandler.js
crolarlibertyva: Turns out that the app_category being passed from index is just the api_category string. But I will relocate the find to index.js.

oauth-proxy/oauthHandlers/authorizeHandler.js
crolarlibertyva: Will make this so and keep Beau's original finding intact.

oauth-proxy/index.js
```diff
@@ -252,6 +252,10 @@ function buildApp(
     service_issuer,
     okta_client
   ) {
+    const app_category = config.routes.categories.find((item) => {
```
beaugrantham-va, line 255: Can this object just be passed into the `buildMetadataForOpenIdConfiguration` function instead of looking it up?

We already use it to set the first two params (`api_category`, and `manage_endpoint`) so it's readily available.

Or does this complicate anything?

oauth-proxy/oauthHandlers/authorizeHandler.js
```diff
@@ -151,4 +168,47 @@ const checkParameters = async (
...
+      {
+        client_id: client_id,
+      },
+      dynamo_clients_table
+    );
+    if (clientInfo.Item) {
+      clientInfo = clientInfo.Item;
+    } else {
+      return {
+        status: 400,
```
beaugrantham-va, line 189: These status values are no longer used by the caller.

oauth-proxy/oauthHandlers/authorizeHandler.js
```diff
@@ -151,4 +168,47 @@ const checkParameters = async (
...
+        error: "invalid_client",
```
beaugrantham-va, line 190: These error names are all the same. Assuming everything here is indicative of an `invalid_client`, does it make sense for the caller to set that?

oauth-proxy/oauthHandlers/authorizeHandler.js
```diff
@@ -151,4 +168,47 @@ const checkParameters = async (
...
+        error_description: "Invalid client for the authorization path",
```
beaugrantham-va, line 191: A client which can't be found in Okta returns the following error_description today:

```
The client specified by the application is not valid.
```

I think a dynamo miss should be the same.

oauth-proxy/index.js
crolarlibertyva: Yes. Will do.

oauth-proxy/oauthHandlers/authorizeHandler.js
crolarlibertyva: Taking these out.

oauth-proxy/oauthHandlers/authorizeHandler.js
crolarlibertyva: This will come out.

beaugrantham-va (dismissed):

lukemarkwordtlibertyits (approved):
looks good

lukemarkwordtlibertyits (approved):
look good!

PR #238, opened 2021-02-17
Api 4620 basic account behavior fix

crolarlibertyva:
- Uses query parameter to get relay state
- More unit tests

### DIFF
--- a/saml-proxy/package.json
+++ b/saml-proxy/package.json
@@ -36,7 +36,7 @@
     "collectCoverage": true,
     "coverageThreshold": {
       "global": {
-        "lines": 77
+        "lines": 80.65
       }
     },
     "transform": {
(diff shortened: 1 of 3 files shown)


bdgrider-va:
I am getting error messages that seem to pertain to relay state when manually exercising this locally. Could be an issue on my end with config, thought.

image

bdgrider-va (dismissed):
All linting and unit tests pass. There appears to be an issue with how the authnRequest/relay state is being handled, though.

crolarlibertyva:
> I am getting error messages that seem to pertain to relay state when manually exercising this locally. Could be an issue on my end with config, though.
> 
> image
> 
> image
> 
> ^ The above is referencing the relayStateQuery, which was undefined, on line 27 of utils.js.

Pushed another change. Had the grab in the wrong spot.

bdgrider-va (approved):
Verified unit tests, normal auth flow with test user 001, and verified before and after basic MHV user behavior (master shows error page, this branch redirects to verify page).

PR #241, opened 2021-02-23
cctest staging

crolarlibertyva:
- Add check for a test with staging for the client credentials flow.


beaugrantham-va (approved):
Approving as a quick fix. This entire if if if block should go away in the long term and just use `$HOST`.

PR #243, opened 2021-03-01
Create Configurable Metadata

lukemarkwordtlibertyits:
# Description

The Oauth-Proxy should have the ability to override an issuer's metadata if necessary.

```json
{
  "api_category": "/override_Endpoints",
  "upstream_issuer": "http://example.com/issuer",
  "manage_endpoint": "http://example.com/account",
  "custom_metadata": {
    "authorization_endpoint": "http://example.custom.com/authorization",
    "token_endpoint": "http://example.custom.com/token",
    "userinfo_endpoint": "http://example.custom.com/userinfo",
    "introspection_endpoint": "http://example.custom.com/introspect",
    "revocation_endpoint": "http://example.custom.com/revoke",
    "jwks_uri": "http://example.custom.com/keys",
    "issuer": "http://example.custom.com/keys",
  }
}
```

**GET Endpoints**

- User info | Used by Issuer Client
- Jwks | Used by Issuer Client

**POST Endpoints**

- authorization | Used by Issuer Client
- token | Used by Issuer Client
- introspect  | Used by Issuer Client
- revocation  | Used by Issuer Client
- issuer | Used by Issuer Client

~**Should the delete grants endpoint be included?**~

~The issuer with this endpoint, is that the oauth-proxy uses the okta-client to perform this logic. The delete grants endpoint for okta is specific to its own API. Other AuthZ servers may not contain the same delete grants endpoint.~

~If this functionality is required, then it may be worth creating a wrapper authz-client class.~

The Delete Grants endpoint is not specified on the OpenId spec, so it will not be taken into consideration.

# TODO

- [x] Create Scripted Tests for this functionality
- [x] Unit Tests
- [x] Implement

# Tests

Since the issuers that will need this functionality are not ready, testing this functionality will be a little tricky.

## JWKS endpoint

```sh
curl $HOST/oauth2/keys
```

**Expected**

- [x] Redirect to the proxied issuer's JWKS endpoint.

## User Info endpoint

```sh
curl $HOST/oauth2/userinfo
```

**Expected**

- [x] Redirect to the proxied issuer's User Info endpoint.

## Post Endpoint tests

- Go through auth follow
- Introspect against the default issuer
  - [x] should return false
- Introspect against the proxied issuer that is used in the introspect swap 
   - [x] should return true
- revoke
- Introspect against the proxied issuer that is used in the introspect swap
  - [x] should return false

## Regression Tests 
- [x] Pass with no changes to config
- [x] Pass when the custom_metadata values consistently point to a different issuer than the upstream_issuer


oauth-proxy/cli.js
```diff
@@ -151,6 +151,53 @@ function processArgs() {
             required: false,
             default: null,
           },
+          custom_metadata: {
+            type: "array",
```
beaugrantham-va, line 155: This isn't an array.

oauth-proxy/cli.js
```diff
@@ -151,6 +151,53 @@ function processArgs() {
...
+            description:
+              "An array of endpoints that can be used to override the default metadata.",
+            required: true,
```
beaugrantham-va, line 158: This should not be required.

crolarlibertyva:
Works great so far! Was able to override all the endpoints to my local keycloak instance.

oauth-proxy/issuer_helper.js
```diff
@@ -0,0 +1,25 @@
+const { Issuer } = require("openid-client");
+
+const buildIssuer = async (serviceConfig) => {
```
crolarlibertyva, line 3: Should we add jsdocs here since it's a new module?

beaugrantham-va (approved):

crolarlibertyva (approved):

PR #244, opened 2021-03-02
Authorize Handler Refactor

lukemarkwordtlibertyits:
# Description

Precursor to creating RFC compliant authorize error handling. Refactor previous error handling to eliminate code smells. Reduce nested if statements. 

# TODO

- [x] Organize Unit Tests by Relationship
- [x] Local vs Auth Client Validation broken into helper method
- [x] checks return objects instead of throwing errors

There are no e2e tests for the authorization endpoint. Add the following:
- [x] Happy Path
- [x] invalid_request missing state | will eventually redirect
- [x] invalid_request missing redirect | notify resource owner

# Tests

- [x] Regression Tests Run
- [x] Unit Tests Pass
## Confirm Happy Path is working.

- Make Authorization request.
- client_store = local
- client_store has redirect `localhost:8080/redirect`

### Expected results

- [x] Redirect to authorization server 

## Confirm Client Redirect Miss is working.

- Make Authorization request.
- client_store = local
- client_store does not have redirect `localhost:8080/redirect`

expected results

- [x] expected json
```json
{
    "error":	"invalid_client",
    "error_description": "The redirect URI specified by the application does not match any of the registered redirect URIs. Erroneous redirect URI: http://localhost:8080/redirect"
}
```

## Client miss in client_store

- Make Authorization request.
- client_store = local
- specified client not in client_store

expected results

- [x] expected json
```json
{
    "error": "invalid_client",
    "error_description": "Invalid client for the authorization path",
}
```

## No Clients in client_store

- Make Authorization request.
- client_store = local
- force dynamo error 

expected results

- [x] expected json
```json
{
    "error": "invalid_client",
    "error_description": "The client specified by the application is not valid."
}
```

## No Okta Redirect URI

- Make Authorization request.
- client_store = null
- redirect uri not registered with okta

expected results

- [x] expected json
```json
{
    "error": "invalid_client",
    "error_description": "The redirect URI specified by the application does not match any of the registered redirect URIs. Erroneous redirect URI: ${client_redirect}"
}
```

## No Okta Client Id

- Make Authorization request.
- client_store = null
- client id not registered with okta

expected results

- [x] expected json
```json
{
    "error": "invalid_client",
    "error_description": "The client specified by the application is not valid.",
}
```

### DIFF
--- a/oauth-proxy/oauthHandlers/authorizeHandler.js
+++ b/oauth-proxy/oauthHandlers/authorizeHandler.js
@@ -19,71 +19,40 @@ const authorizeHandler = async (
 ) => {
   loginBegin.inc();
   const { state, client_id, aud, redirect_uri: client_redirect } = req.query;
-  try {
-    await checkParameters(
-      state,
-      aud,
-      issuer,
-      logger,
-      oktaClient,
-      client_redirect
-    );
-  } catch (err) {
-    if (err.status == 500) {
-      return next(err);
-    }
-    res.status(err.status).json({
-      error: err.error,
-      error_description: err.error_description,
+
+  let missingParameters = await checkParameters(
+    state,
+    aud,
+    issuer,
+    logger,
+    oktaClient,
+    client_redirect
+  );
+
+  if (missingParameters) {
+    res.status(missingParameters.status).json({
+      error: missingParameters.error,
+      error_description: missingParameters.error_description,
     });
     return next();
   }
 
-  if (app_category.client_store && app_category.client_store === "local") {
-    const errInfo = await localValidateClient(
-      logger,
-      client_id,
-      client_redirect,
-      dynamoClient,
-      dynamo_clients_table
-    );
-    if (errInfo) {
-      res.status(400).json({
-        error: "invalid_client",
-        error_description: errInfo.error_description,
-      });
-      return next();
-    }
-  } else {
-    try {
-      const oktaApp = await oktaClient.getApplication(client_id);
-      if (
-        oktaApp.settings.oauthClient.redirect_uris.indexOf(client_redirect) ===
-        -1
-      ) {
-        res.status(400).json({
-          error: "invalid_client",
-          error_description:
-            "The redirect URI specified by the application does not match any of the " +
-            `registered redirect URIs. Erroneous redirect URI: ${client_redirect}`,
-        });
-        return next();
-      }
-    } catch (error) {
-      // This error is unrecoverable because we would be unable to verify
-      // that we are redirecting to a whitelisted client url
-      logger.error(
-        "Unrecoverable error: could not get the Okta client app",
-        error
-      );
+  let validationError = await validateClient(
+    logger,
+    client_id,
+    client_redirect,
+    dynamoClient,
+    dynamo_clients_table,
+    oktaClient,
+    app_category
+  );
 
-      res.status(400).json({
-        error: "invalid_client",
-        error_description:
-          "The client specified by the application is not valid.",
-      });
-      return next();
-    }
+  if (validationError) {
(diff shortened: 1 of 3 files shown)


lukemarkwordtlibertyits:
A small change in functionality in how the authorization handler [Handles Authorization Sever Errors](https://github.com/department-of-veterans-affairs/vets-saml-proxy/blob/master/oauth-proxy/oauthHandlers/authorizeHandler.js#L32). It use to return `mext(err)` now it allows the error to bubble up to the index.js file. Also the handler use to [create a generic 500](https://github.com/department-of-veterans-affairs/vets-saml-proxy/blob/master/oauth-proxy/oauthHandlers/authorizeHandler.js#L170) for these types of errors, but not it just uses the error the authorization server throws. I believe this is a better pattern, but would like to get a second opinion.

oauth-proxy/oauthHandlers/authorizeHandler.js
```diff
@@ -198,12 +198,16 @@ const localValidateClient = async (
       clientInfo = clientInfo.Item;
     } else {
       return {
+        status: 400,
```
beaugrantham-va, line 201: These two fields (`status`, `error`) were removed in a previous PR for a couple reasons.

1. The `error` is the same for _all_ scenarios. Someone calling `*ValidateClient` should assume that a validation error is indicative of an `invalid_client`.
2. These functions should not be aware of HTTP status codes. They don't know anything about the caller or what a validation error means in the big picture. The caller should decide this.

oauth-proxy/oauthHandlers/authorizeHandler.js
```diff
@@ -212,10 +216,49 @@ const localValidateClient = async (
...
     return {
+      status: 400,
+      error: "invalid_client",
       error_description:
         "The client specified by the application is not valid.",
     };
   }
 };
 
+const serverValidateClient = async (
```
beaugrantham-va, line 227: New functions should get documentation.

oauth-proxy/oauthHandlers/authorizeHandler.js
```diff
@@ -178,6 +150,34 @@ const checkParameters = async (
       });
     }
   }
+  return null;
+};
+
+const validateClient = async (
```
beaugrantham-va, line 156: Add documentation.

oauth-proxy/oauthHandlers/authorizeHandler.js
```diff
@@ -178,6 +150,34 @@ const checkParameters = async (
       });
     }
   }
+  return null;
```
beaugrantham-va, line 153: This is a minor difference from the other validation/check functions, which return `undefined` (implicitly).

Whatever we return (or don't return) should be consistent. You might even consider returning an object in all cases `{ valid: boolean, error: optional, error_description: optional }` for the caller to clearly check `if (!something.valid) {...}` but that might be scope creep.

It just needs to be consistent and documented.

oauth-proxy/oauthHandlers/authorizeHandler.js
lukemarkwordtlibertyits: That is a good point

oauth-proxy/oauthHandlers/authorizeHandler.js
lukemarkwordtlibertyits: They are not publicaly exposed. Just helper functions. Should they still get documented?

beaugrantham-va: I'd vote yes, since there are multiple `*ValidateClient` functions... it might be unclear to a future developer which one to call or what they expect to get back.

beaugrantham-va (approved):

PR #245, opened 2021-03-03
RFC compliant authorize error handling

lukemarkwordtlibertyits:
# Description

Authorize Handler should return RFC compliant errors,

```
If the request fails due to a missing, invalid, or mismatching
redirection URI, or if the client identifier is missing or invalid,
the authorization server SHOULD inform the resource owner of the
error and MUST NOT automatically redirect the user-agent to the
invalid redirection URI.

If the resource owner denies the access request or if the request
fails for reasons other than a missing or invalid redirection URI,
the authorization server informs the client by adding the following
parameters to the query component of the redirection URI using the
"application/x-www-form-urlencoded" format, per Appendix B:
...
```

In the case of the Oauth-Proxy

- Issues with the redirect_uri can continue to result in an error page rendered by the oauth-proxy.
- Other authorization errors should be sent to the redirect_uri. ex.
```http
https://example-app.com/cb?error=invalid_request&error_description=blah blah blah
```

## Different error handling scenarios

**invalid_request**
- no state

**invalid_request cannot be sent back to redirect uri**
- no Redirect URI
- Erronous redirect local | was previously invalid_client
- Erronous redirect Okta | was previously invalid_client

**server_error Oauth Proxy**
- could not pull from dynamodb | returns unauthorized_client
- could not save redirect | bubbles up

**server_error Okta**
- could not get auth server | bubbles up

**unauthorized_client**
- bad client local | does not redirect
- could not get okta app | does not redirect

# TODO

Modify unit tests to check for errors on the following scenarios:
- [x] no state
- [x] could not pull from dynamodb
- [x] could not save redirect
- [x] could not get auth server
- [x] could not get okta app
- [x] bad client 
- [x] add error and error_description checks on 400 tests

Regression Tests
- [x] Add integration test for no Redirect URI
- [x] Modify no state integration test to check for redirect

Integration Tests
- [x] modify missing state integration test to check for redirect

# Tests

## Regression tests pass.
**Happy Path**
- [x]  Pass

**No Redirect**
- [x] Pass
```json
{
  "error": "invalid_request",
  "error_description" : "There was no redirect URI specified by the application."
}
```
was previously invalid_client

**No State**
- [x] Pass
```json
{
  "error": "invalid_request",
  "error_description" : "State parameter required"
}
```
## Local Happy Path
- [x] Redirect to saml-proxy

## Erroneous Redirect
**local**
- [x] 400 Error
- [x] json
```json
{
  "error": "invalid_request ",
  "error_description": "The redirect URI specified by the application does not match any of the registered redirect URIs. Erroneous redirect URI: {URI}"
}
```
**auth server**
- [x] 400 Error
- [x] json
```json
{
  "error": "invalid_request ",
  "error_description": "The redirect URI specified by the application does not match any of the registered redirect URIs. Erroneous redirect URI: {URI}"
}
```

This use to return a invalid_client. Should it still?

## Server Errors
**Could not pull from dynamo**
- [x] 400
- [x] error = unauthorized_client
- [x] error_description = "The client specified by the application is not valid."

**Could not save to dynamo**
- [x] bubbles up.

**Could not get auth server when checking for aud**
- [x] bubbles up.

## Invalid Client

**Could not get okta app**
- [x] 400
- [x] error = unauthorized_client
- [x] error_description = "The client specified by the application is not valid."

**Bad Client Local** 
- [x] 400
- [x] error = unauthorized_client
- [x] error_description = "The client specified by the application is not valid."

### DIFF
--- a/oauth-proxy/oauthHandlers/authorizeHandler.js
+++ b/oauth-proxy/oauthHandlers/authorizeHandler.js
@@ -123,7 +123,7 @@ const authorizeHandler = async (
 /**
  * Checks for valid authorization parameters.
  *
- * @returns {Promise<{valid: boolean, error_description: string, error: string}>}
+ * @returns {Promise<{valid: boolean, error?: string, error_description?: string}>}
  */
 const checkParameters = async (state, aud, issuer, logger, oktaClient) => {
   if (!state) {
@@ -165,7 +165,7 @@ const checkParameters = async (state, aud, issuer, logger, oktaClient) => {
 /**
  * Checks for authorization server or local database for valid client.
  *
- * @returns {Promise<{valid: boolean, error_description: string}>}
+ * @returns {Promise<{valid: boolean, error?: string, error_description?: string}>}
  */
 const validateClient = async (
   logger,
@@ -206,7 +206,7 @@ const validateClient = async (
 /**
  * Checks for authorization local database for valid client.
  *
- * @returns {Promise<{valid: boolean, error_description: string}>}
+ * @returns {Promise<{valid: boolean, error?: string, error_description?: string}>}
  */
 const localValidateClient = async (
   logger,
@@ -256,7 +256,7 @@ const localValidateClient = async (
 /**
  * Checks for authorization server for valid client.
  *
- * @returns {Promise<{valid: boolean, error_description: string}>}
+ * @returns {Promise<{valid: boolean, error?: string, error_description?: string}>}
  */
 const serverValidateClient = async (
   oktaClient,
@@ -298,6 +298,8 @@ const serverValidateClient = async (
 
 /**
  * Builds errors to be sent to client's redirect uri.
+ *
+ * @returns {module:url.URL}
  */
 const buildRedirectErrorUri = (err, redirect_uri) => {
   let uri = new URL(redirect_uri);
@@ -305,4 +307,5 @@ const buildRedirectErrorUri = (err, redirect_uri) => {
   uri.searchParams.append("error_description", err.error_description);
   return uri;
 };
+
 module.exports = authorizeHandler;
--- a/oauth-proxy/oauthHandlers/authorizeHandler.js
+++ b/oauth-proxy/oauthHandlers/authorizeHandler.js
@@ -35,37 +35,8 @@ const authorizeHandler = async (
   res,
   next
 ) => {
-  // implement rfc compliant error handling ...
   loginBegin.inc();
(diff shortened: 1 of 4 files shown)


oauth-proxy/oauthHandlers/authorizeHandler.js
```diff
@@ -232,13 +248,17 @@ const localValidateClient = async (
     } else {
       return {
         valid: false,
+        redirect: true,
```
beaugrantham-va, line 251: If a client is not retrievable, there is no way we can validate the redirect URI. This should be false.

oauth-proxy/oauthHandlers/authorizeHandler.js
```diff
@@ -279,6 +301,8 @@ const serverValidateClient = async (
 
     return {
       valid: false,
+      redirect: true,
```
beaugrantham-va, line 304: If a client is not retrievable, there is no way we can validate the redirect URI. This should be false.

oauth-proxy/oauthHandlers/authorizeHandler.js
```diff
@@ -38,37 +38,58 @@ const authorizeHandler = async (
...
+    client_redirect,
+    dynamoClient,
+    dynamo_clients_table,
     oktaClient,
-    client_redirect
+    app_category
   );
 
-  if (!missingParameters.valid) {
+  if (!validationError.valid && validationError.redirect) {
```
beaugrantham-va, line 51: I can't think of any scenario where a redirect would be permitted if there is a problem with the client or redirect URI.

oauth-proxy/oauthHandlers/authorizeHandler.js
```diff
@@ -38,37 +38,58 @@ const authorizeHandler = async (
...
-    client_redirect,
-    dynamoClient,
-    dynamo_clients_table,
-    oktaClient,
-    app_category
+    oktaClient
   );
 
-  if (!validationError.valid) {
+  if (!missingParameters.valid && missingParameters.redirect) {
```
beaugrantham-va, line 78: I think any error once we validate client and redirect URI should be a redirect. I don't see any scenarios where we wouldn't (other than a 500, which gets bubbled up).

beaugrantham-va (changes requested):

crolarlibertyva (approved):

beaugrantham-va (approved):

PR #250, opened 2021-03-10
API-5384 support for custom certs in docker image

bdgrider-va:
These changes allow us to add a new certificate to the two different docker images that are produced for our two different deployment models. To support Jenkins deployments, the ansible build script was modified to copy the fwdproxy self-signed cert to the expected location /opt/samlproxy. The Dockerfile then appends that to the system certs located in /ca-certificates/ca-certs.pem. For our ECS deployments, the fwd proxy self-signed cert is loaded in parameter store, then copied to the expected location /opt/samlproxy by _____. The DockerfileFG then appends the cert to the system certs located in /ca-certificates/ca-certs.pem.

Related PRs:
https://github.com/department-of-veterans-affairs/devops/pull/8704/files

Test steps:

1) Manually put the cert located here https://github.com/department-of-veterans-affairs/devops/blob/master/ansible/deployment/config/fwdproxy/self.crt in the root of the vets-saml-proxy repo as "fwdproxy.crt"

2) Build the oauth-proxy docker image from the root of vets-saml-proxy (you may have to run make login before this step) as TARGET deploy
```
make build IMAGE=oauth-proxy TARGET=deploy
```

2) Run the docker container 
```
docker run --rm --name testcontainer \
  -v ${PWD}/oauth-proxy/dev-config.json:/oauth-proxy/config.json \
  {IMAGE_NAME}
```

3) Exec into the running container and verify that the certificate was appended to the end of the certificate chain in /ca-certificates/ca-certs.pem.
```
docker exec -it testcontainer /bin/bash
cat /ca-certificates/ca-certs.pem
```

### DIFF
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -65,7 +65,7 @@ pipeline {
           dir "."
           filename "oauth-proxy/Dockerfile"
           label 'vetsgov-general-purpose'
-          additionalBuildArgs '--pull'
+          additionalBuildArgs '--target base --pull'
         }
       }
 
@@ -86,7 +86,7 @@ pipeline {
           dir "."
           filename "oauth-proxy/Dockerfile"
           label 'vetsgov-general-purpose'
-          additionalBuildArgs '--pull'
+          additionalBuildArgs '--target base --pull'
         }
       }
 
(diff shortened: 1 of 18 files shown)


beaugrantham-va:
Is there a corresponding PR to populate `$CUSTOM_CERTS` for both the AMI and ECS deployments?

PR #254, opened 2021-03-11
Saves access-token and issuer on Token Request

lukemarkwordtlibertyits:
# Description

Add Access Token and ISS to OAuthRequestsV2 on Token requests.

# TODO

- [x] Create PR to Add `oauth_access_token_index` to `OAuthRequestsV2`
    https://github.com/department-of-veterans-affairs/lighthouse-devops-support/issues/280
    - [x] Add `oauth_access_token_index` to `OAuthRequestsV2` in `dynamo_schema.js`
    **note** this index is not required to store the access_token, but will be important for [API-5575](https://vajira.max.gov/browse/API-5575)
- [x] Add logic to store access_token in the tokenHandlerClient.
- [x] In SaveDocumentByState create a common payload instantiation to reduce repeated code.
- [x] Unit Tests
- [x] Manual Tests

# Tests

## Happy Path Okta no offline_access

**Test Steps**

- Go through Oauth flow on default issuer with no offline access scope. Proves that token flow saves document with no refresh_token,

**Expected Results**

```sh
aws dynamodb scan --table-name OAuthRequestsV2 --endpoint-url http://localhost:8000
```

should return following fields

- [x] code
- [x] internal_state
- [x] redirect_uri
- [x] state
- [x] expires_on
- [x] access_token
- [x] iss

### Happy Path

**Test Steps**

- Go through Oauth flow on default issuer.

**Expected Results**

```sh
aws dynamodb scan --table-name OAuthRequestsV2 --endpoint-url http://localhost:8000
```

should return following fields

- [x] code
- [x] internal_state
- [x] redirect_uri
- [x] state
- [x] expires_on
- [x] access_token
- [x] refresh_token
- [x] iss

## Refresh

**Test Steps**

- Go through refresh flow on default issuer.

**Expected Results**

```sh
aws dynamodb scan --table-name OAuthRequestsV2 --endpoint-url http://localhost:8000
```

should return following fields

- [x] code
- [x] internal_state
- [x] redirect_uri
- [x] state
- [x] expires_on
- [x] access_token
- [x] refresh_token
- [x] iss

## Refresh Pull Document from OAuthRequestsV1

**Test Steps**

- Go through Token Flow
- Delete Token Item in OAuthRequestsV2
- Inject Code, State and Refresh_Token into OAuthRequests
- Go through refresh token flow.

**Expected Results**

```sh
aws dynamodb scan --table-name OAuthRequestsV2 --endpoint-url http://localhost:8080
```

should return following fields

- [x] code
- [x] internal_state
- [x] redirect_uri
- [x] state
- [x] expires_on
- [x] access_token
- [x] refresh_token
- [x] iss

~~Refresh Pull Document from OAuthRequestsV1 no Access Token~~

Does not make sense...

### DIFF
--- a/oauth-proxy/README.md
+++ b/oauth-proxy/README.md
@@ -21,7 +21,7 @@ Set up Dynamo DB locally
 The following dev-config value needs to be set `"dynamo_local": "localhost:8000",`
 
 ```
-docker run -p 8000:8000 amazon/dynamodb-local -jar DynamoDBLocal.jar -sharedDb -inMemory
+docker run -d -p 8000:8000 amazon/dynamodb-local -jar DynamoDBLocal.jar -sharedDb -inMemory
 node dynamo_schema.js --local=true
 ```
 
--- a/oauth-proxy/README.md
+++ b/oauth-proxy/README.md
@@ -21,7 +21,7 @@ Set up Dynamo DB locally
 The following dev-config value needs to be set `"dynamo_local": "localhost:8000",`
 
 ```
-docker run -d -p 8000:8000 amazon/dynamodb-local -jar DynamoDBLocal.jar -sharedDb -inMemory
+docker run -p 8000:8000 amazon/dynamodb-local -jar DynamoDBLocal.jar -sharedDb -inMemory
 node dynamo_schema.js --local=true
 ```
 
--- a/oauth-proxy/oauthHandlers/tokenHandlerStrategyClasses/saveDocumentStrategies/saveDocumentStateStrategy.js
+++ b/oauth-proxy/oauthHandlers/tokenHandlerStrategyClasses/saveDocumentStrategies/saveDocumentStateStrategy.js
@@ -48,17 +48,13 @@ class SaveDocumentStateStrategy {
               this.config.hmac_secret
             ),
             iss: this.issuer,
-          };
-          if (tokens.refresh_token) {
-            payload.refresh_token = hashString(
+            refresh_token: hashString(
               tokens.refresh_token,
               this.config.hmac_secret
-            );
-            payload.expires_on =
-              Math.round(Date.now() / 1000) + 60 * 60 * 24 * 42;
-          } else {
-            payload.expires_on = tokens.expires_at;
-          }
+            ),
+            expires_on: Math.round(Date.now() / 1000) + 60 * 60 * 24 * 42,
+          };
+
           if (document.launch) {
             payload.launch = document.launch;
           }
--- a/oauth-proxy/oauthHandlers/tokenHandlerStrategyClasses/saveDocumentStrategies/saveDocumentStateStrategy.js
+++ b/oauth-proxy/oauthHandlers/tokenHandlerStrategyClasses/saveDocumentStrategies/saveDocumentStateStrategy.js
@@ -3,12 +3,12 @@ const jwtDecode = require("jwt-decode");
 const { v4: uuidv4 } = require("uuid");
 
 class SaveDocumentStateStrategy {
-  constructor(req, logger, dynamoClient, config, issuer_client) {
+  constructor(req, logger, dynamoClient, config, issuer) {
     this.req = req;
     this.logger = logger;
     this.dynamoClient = dynamoClient;
     this.config = config;
-    this.issuer_client = issuer_client;
+    this.issuer = issuer;
   }
   async saveDocumentToDynamo(document, tokens) {
     try {
@@ -19,7 +19,7 @@ class SaveDocumentStateStrategy {
               tokens.access_token,
               this.config.hmac_secret
             ),
-            iss: this.issuer_client.issuer,
+            iss: this.issuer,
           };
 
           if (tokens.refresh_token) {
@@ -47,7 +47,7 @@ class SaveDocumentStateStrategy {
               tokens.access_token,
               this.config.hmac_secret
             ),
-            iss: this.issuer_client.issuer,
+            iss: this.issuer,
           };
           if (tokens.refresh_token) {
             payload.refresh_token = hashString(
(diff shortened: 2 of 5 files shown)


crolarlibertyva (approved):

oauth-proxy/oauthHandlers/tokenHandlerStrategyClasses/saveDocumentStrategies/saveDocumentStateStrategy.js
```diff
@@ -20,6 +21,11 @@ class SaveDocumentStateStrategy {
                 tokens.refresh_token,
                 this.config.hmac_secret
               ),
+              access_token: hashString(
```
beaugrantham-va, line 24: This will only store access tokens when a refresh token is present.

We'll need this to work without a refresh token.

beaugrantham-va (changes requested):

oauth-proxy/oauthHandlers/tokenHandlerStrategyClasses/saveDocumentStrategies/saveDocumentStateStrategy.js
lukemarkwordtlibertyits: I see what you are saying. Now it is necessary to check for the access_token instead of the refresh token. There is no situation where there will be an access_token with no refresh_token, but there are situations where the opposite is true. My changes will reflect this.

oauth-proxy/oauthHandlers/tokenHandlerStrategyClasses/saveDocumentStrategies/saveDocumentStateStrategy.js
```diff
@@ -3,26 +3,37 @@ const jwtDecode = require("jwt-decode");
...
+      if (document.state && tokens.access_token) {
         if (document.internal_state) {
+          let updated_document = {
+            access_token: hashString(
+              tokens.access_token,
+              this.config.hmac_secret
+            ),
+            iss: this.issuer_client.issuer,
+            // 42 days
+            expires_on: Math.round(Date.now() / 1000) + 60 * 60 * 24 * 42,
```
beaugrantham-va, line 24: From the story:

> For cases where there is no refresh_token, update TTL field based on expires value in token response (e.g. 1 hour).

beaugrantham-va (changes requested):

oauth-proxy/oauthHandlers/tokenHandlerStrategyClasses/saveDocumentStrategies/saveDocumentStateStrategy.js
crolarlibertyva, line 24: @beaugrantham-va : Should this be set to the exp if there's no refresh token?

oauth-proxy/oauthHandlers/tokenHandlerStrategyClasses/saveDocumentStrategies/saveDocumentStateStrategy.js
```diff
@@ -34,6 +47,11 @@ class SaveDocumentStateStrategy {
               tokens.refresh_token,
               this.config.hmac_secret
             ),
+            access_token: hashString(
+              tokens.access_token,
+              this.config.hmac_secret
+            ),
+            iss: this.issuer_client.issuer,
             // 42 days
             expires_on: Math.round(Date.now() / 1000) + 60 * 60 * 24 * 42,
```
crolarlibertyva, line 56: This should be the same as in the if condition.

oauth-proxy/oauthHandlers/tokenHandlerStrategyClasses/saveDocumentStrategies/saveDocumentStateStrategy.js
```diff
@@ -3,40 +3,62 @@ const jwtDecode = require("jwt-decode");
 const { v4: uuidv4 } = require("uuid");
 
 class SaveDocumentStateStrategy {
-  constructor(req, logger, dynamoClient, config) {
+  constructor(req, logger, dynamoClient, config, issuer_client) {
     this.req = req;
     this.logger = logger;
     this.dynamoClient = dynamoClient;
     this.config = config;
+    this.issuer_client = issuer_client;
```
beaugrantham-va, line 11: Is there benefit in passing this full object vs just the raw issuer that we need?

oauth-proxy/README.md
```diff
@@ -21,7 +21,7 @@ Set up Dynamo DB locally
 The following dev-config value needs to be set `"dynamo_local": "localhost:8000",`
 
 ```
-docker run -p 8000:8000 amazon/dynamodb-local -jar DynamoDBLocal.jar -sharedDb -inMemory
+docker run -d -p 8000:8000 amazon/dynamodb-local -jar DynamoDBLocal.jar -sharedDb -inMemory
```
lukemarkwordtlibertyits, line 24: I always add this into the command so it may as well go into the ReadMe

oauth-proxy/oauthHandlers/tokenHandlerStrategyClasses/saveDocumentStrategies/saveDocumentStateStrategy.js
lukemarkwordtlibertyits, line 55: Regarding Larry's previous comment about the expires_on logic should take into account a lack of a refresh token down here. I do not believe that is true.  The only way A document from OAuthRequests(V1) could make it here is if it had a refresh token and was created before the internal state logic was added.

crolarlibertyva (approved):

oauth-proxy/oauthHandlers/tokenHandlerStrategyClasses/saveDocumentStrategies/saveDocumentStateStrategy.js
dbrown-va: Agreed, that scenario should only occur with the Refresh Token flow at this point.

dbrown-va (approved):
Changes look good.

The access token as a secondary index looks good in the local dynamo_schema.js.  Since we aren't querying on it yet, I don't think that devops support request would technically be a blocker.

beaugrantham-va (approved):

PR #256, opened 2021-03-12
Grab Scopes from Token Objects instead of Access Token

lukemarkwordtlibertyits:
- [x] Remove reference to decoded Access token
  - tokenHandlerClient
  - saveDocumentStateStrategy
- [x] Unit tests work
- [x] Regression tests work

### DIFF
--- a/oauth-proxy/oauthHandlers/tokenHandlerStrategyClasses/saveDocumentStrategies/saveDocumentStateStrategy.js
+++ b/oauth-proxy/oauthHandlers/tokenHandlerStrategyClasses/saveDocumentStrategies/saveDocumentStateStrategy.js
@@ -58,7 +58,8 @@ class SaveDocumentStateStrategy {
     try {
       if (document.launch && tokens.access_token) {
         let decodedToken = jwtDecode(tokens.access_token);
-        if (decodedToken.scp.includes("launch")) {
+        let scp = decodedToken.scp;
+        if (scp.includes("launch")) {
           let launch = document.launch;
           let accessToken = hashString(
             tokens.access_token,
--- a/oauth-proxy/oauthHandlers/tokenHandlerStrategyClasses/saveDocumentStrategies/saveDocumentStateStrategy.js
+++ b/oauth-proxy/oauthHandlers/tokenHandlerStrategyClasses/saveDocumentStrategies/saveDocumentStateStrategy.js
@@ -58,7 +58,7 @@ class SaveDocumentStateStrategy {
     try {
       if (document.launch && tokens.access_token) {
         let decodedToken = jwtDecode(tokens.access_token);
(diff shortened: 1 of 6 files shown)


oauth-proxy/oauthHandlers/tokenHandlerStrategyClasses/saveDocumentStrategies/saveDocumentStateStrategy.js
```diff
@@ -58,7 +58,7 @@ class SaveDocumentStateStrategy {
     try {
       if (document.launch && tokens.access_token) {
         let decodedToken = jwtDecode(tokens.access_token);
-        if (decodedToken.scp.includes("launch")) {
+        if (tokens.scope && tokens.scope.includes("launch")) {
```
beaugrantham-va, line 61: `decodedToken.scp` is an array while `tokens.scope` is a string.

Using `.includes` will result in different behavior here. It needs to match `launch` exactly.

oauth-proxy/oauthHandlers/tokenHandlerStrategyClasses/tokenHandlerClient.js
```diff
@@ -68,14 +68,15 @@ class TokenHandlerClient {
...
+
+    if (tokens && tokens.scope) {
+      if (tokens.scope.includes("launch/patient")) {
         let patient = await this.getPatientInfoStrategy.createPatientInfo(
           tokens,
           decoded
         );
         responseBody["patient"] = patient;
-      } else if (decoded.scp.includes("launch") && launch) {
+      } else if (tokens.scope.includes("launch") && launch) {
```
beaugrantham-va, line 79: `decoded.scp` is an array while `tokens.scope` is a string.

Using `.includes` will result in different behavior here. It needs to match `launch` exactly.

beaugrantham-va (changes requested):

lukemarkwordtlibertyits:
To test Regex:

```js
let arr = "launch patient/launch".split(/(\s)/);

arr.forEach(element => console.log(element));
```

https://jsfiddle.net/vh93u851/4/

beaugrantham-va:
> To test Regex:
> 
> ```js
> let arr = "launch patient/launch".split(/(\s)/);
> 
> arr.forEach(element => console.log(element));
> ```
> 
> https://jsfiddle.net/vh93u851/4/

Using a capture group in this regular expression populates the array with unnecessary elements (size = 3).

> If separator is a regular expression with capturing parentheses, then each time separator matches, the results (including any undefined results) of the capturing parentheses are spliced into the output array.

We should drop the capture group or use a string literal (`' '`) to reduce the number of elements to search through (size = 2).

beaugrantham-va (approved):

PR #260, opened 2021-03-16
API-5571 audience config

crolarlibertyva:
- Adds the audience setting in the config for each app category

Config PRs
* https://github.com/department-of-veterans-affairs/lighthouse-oauth-proxy-configs/pull/24
* https://github.com/department-of-veterans-affairs/devops/pull/8765


beaugrantham-va (changes requested):
Needs corresponding devops config PR.

beaugrantham-va (changes requested):
Linting fails.

lukemarkwordtlibertyits (commented):
Looks good, my only comment is that the AuthorizationHandler grabs the Audience from the authorizationServer. This will likely cause issues with the new Issuers. It may be worth using the audience from the config in this check. Also this will be less costly!

oauth-proxy/cli.js
```diff
@@ -157,6 +157,11 @@ function processArgs() {
             required: false,
             default: false,
           },
+          audience: {
+            description: "The audience for the app catetory.",
```
beaugrantham-va, line 161: s/catetory/category

oauth-proxy/oauthHandlers/tokenHandlerStrategyClasses/getPatientInfoStrategies/getPatientInfoFromValidateEndpointStrategy.js
```diff
@@ -1,15 +1,16 @@
 const { rethrowIfRuntimeError } = require("../../../utils");
 class GetPatientInfoFromValidateEndpointStrategy {
-  constructor(validateToken, logger) {
+  constructor(validateToken, logger, audience) {
     this.logger = logger;
     this.validateToken = validateToken;
+    this.audience = audience;
   }
-  async createPatientInfo(tokens, decoded) {
```
beaugrantham-va, line 7: Anywhere that calls this function needs to be updated to match the new signature.

beaugrantham-va, line 7: A side effect of changing this signature is that you'll need to change the signature of the other classes that implement the same "interface" - getPatientInfoFromLaunchStrategy.js

beaugrantham-va (changes requested):
Can you pull the latest master?

oauth-proxy/tests/TokenHandlerTests/tokenHandler.test.js
```diff
@@ -39,7 +40,7 @@ beforeEach(() => {
   });
   validateToken = buildValidateToken({});
   staticTokens = new Map();
-  next = jest.fn();
+  (audience = "https://sandbox-api.va.gov/services/fhir"), (next = jest.fn());
```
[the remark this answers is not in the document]
bdgrider-va: I think the linter might have done this? Changing it back.

oauth-proxy/oauthHandlers/tokenHandlerStrategyClasses/getPatientInfoStrategies/getPatientInfoFromLaunchStrategy.js
```diff
@@ -3,7 +3,7 @@ class GetPatientInfoFromLaunchStrategy {
     this.req = req;
   }
   // eslint-disable-next-line no-unused-vars
```
beaugrantham-va, line 5: This eslint directive can be removed now.

beaugrantham-va (changes requested):
Still need to pull in the latest master

oauth-proxy/oauthHandlers/tokenHandlerStrategyClasses/getPatientInfoStrategies/getPatientInfoFromLaunchStrategy.js
bdgrider-va: It's still needed for the tokens parameter.

beaugrantham-va: You're totally right. My bad.

lukemarkwordtlibertyits (changes requested):
This is kindof a tough one to spot, but I am not sure that the config audience ever gets passed through. See this log on the Authorization Handler.

image

If the audience was passed then there would also be an "expected" field in that message.

I think the same is true in the refresh flow.

bdgrider-va:
> This is kindof a tough one to spot, but I am not sure that the config audience ever gets passed through. See this log on the Authorization Handler.
> 
> image
> 
> If the audience was passed then there would also be an "expected" field in that message.
> 
> I think the same is true in the refresh flow.

Good catch, just fixed this.

image

oauth-proxy/oauthHandlers/authorizeHandler.js
```diff
@@ -125,7 +120,7 @@ const authorizeHandler = async (
  *
  * @returns {Promise<{valid: boolean, error?: string, error_description?: string}>}
  */
-const checkParameters = async (state, aud, issuer, logger, oktaClient) => {
+const checkParameters = async (state, aud, audience, logger) => {
```
beaugrantham-va, line 123: These param names (`aud`, `audience`) need to be more descriptive and needs `@param` documentation.

beaugrantham-va (changes requested):

oauth-proxy/oauthHandlers/authorizeHandler.js
```diff
@@ -28,6 +28,7 @@ const authorizeHandler = async (
   oktaClient,
   slugHelper,
   app_category,
+  audience,
```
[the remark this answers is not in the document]
beaugrantham-va: If not, you'll need to update the jsdoc.

oauth-proxy/oauthHandlers/authorizeHandler.js
```diff
@@ -136,26 +131,11 @@ const checkParameters = async (state, aud, issuer, logger, oktaClient) => {
...
-      .then((res) => {
-        serverAudiences = res.audiences;
-      })
-      .catch(() => {
-        logger.error("Unable to get the authorization server.");
-        throw { status: 500 };
-      });
-
-    if (!serverAudiences.includes(aud)) {
+    if (!(aud === audience)) {
```
beaugrantham-va, line 134: Is there a difference between `!(aud === audience)` and `aud !== audience`? Not saying I know... genuinely curious.

bdgrider-va: It appears that they are both "identity" comparisons, where both type and value are evaluated. So
"2"!==2 and !("2"===2) both evaluate to true

image

oauth-proxy/oauthHandlers/authorizeHandler.js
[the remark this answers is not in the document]
bdgrider-va: It does appear that is the case, removing the audience parameter now.

oauth-proxy/oauthHandlers/authorizeHandler.js
lukemarkwordtlibertyits: If all things are equal than I think `!==` is preferable. To me its easier to read.

oauth-proxy/oauthHandlers/tokenHandlerStrategyClasses/getPatientInfoStrategies/getPatientInfoFromValidateEndpointStrategy.js
```diff
@@ -1,15 +1,16 @@
 const { rethrowIfRuntimeError } = require("../../../utils");
 class GetPatientInfoFromValidateEndpointStrategy {
-  constructor(validateToken, logger) {
+  constructor(validateToken, logger, audience) {
     this.logger = logger;
     this.validateToken = validateToken;
+    this.audience = audience;
```
lukemarkwordtlibertyits, line 6: I added a debugging line in this class to check that the audience parameter was being passed in. It is `undefined` my setup is pointing at the v1 validation endpoint so I am not sure why it is passing, still if this is a field of this class, it should be populated.

May be worth looking into whether this field is actually necessary any more?

lukemarkwordtlibertyits (changes requested):

oauth-proxy/oauthHandlers/tokenHandlerStrategyClasses/getPatientInfoStrategies/getPatientInfoFromValidateEndpointStrategy.js
bdgrider-va: Just fixed this. I had missed a second instantiation of the GetPatientInfoFromValidateEndpointStrategy class in the tokenHandlerClientBuidler.

lukemarkwordtlibertyits (dismissed):
Looks good!

beaugrantham-va (approved):

PR #262, opened 2021-03-18
grab exp from token object

lukemarkwordtlibertyits:
# Description

To support SSOi, the token expiration stored for code and refresh flows should be grabbed from the token object.

# Questions

~~There is a launch flow for both saveDocumentLaunchStrategy and saveDocumentStateStrategy. The client_credentials access token does not have an `expires_at` or `expires_on` field. There are also cases where this is true for regular token objects reaching the saveDocumentStateStrategy launch block; however, token objects almost always has the expires_in field, which is a long that tell the duration of time when a token will expire. For example a token may expire in 1 hour.~~ 

~~A couple things to consider:~~

- ~~If we intend to save SSOi tokens in the LaunchContext table it may be best to set the expires_at field as `now + expires_in`.~~
- ~~If we do not intend to save SSOi okens in the LaunchContext table it may be best to continue the use of the access_token's exp field because ideally this field should represent the access_tokens expiration and not the refresh token's.~~

# TODO

- [x] saveDocumentStateStrategy should store the expiration as `now + expires_on`.
- [x] saveDocumentLaunchStrategy should store the expiration as `now + expires_on`.

# Tests

## Happy Path Okta no offline_access

**Test Steps**

- Go through Oauth flow on default issuer with no offline access scope. Proves that token flow saves document with no refresh_token,

**Expected Results**

```sh
aws dynamodb scan --table-name OAuthRequestsV2 --endpoint-url http://localhost:8000
```

should return following fields

- [x] code
- [x] internal_state
- [x] redirect_uri
- [x] state
- [x] expires_on | should match the token object
- [x] access_token
- [x] iss

### Happy Path ~~With Launch~~

**Test Steps**

- Go through Oauth flow on default issuer.

**Expected Results**

```sh
aws dynamodb scan --table-name OAuthRequestsV2 --endpoint-url http://localhost:8000
```

should return following fields

- [x] code
- [x] internal_state
- [x] redirect_uri
- [x] state
- [x] expires_on | 42 days
- [x] access_token
- [x] refresh_token
- [x] iss

It does not appear that there is an issuer configured to work with the launch scope at the moment.

### SSOe

**Test Steps**

- Go through the SSOe Flow

**Expected Results**

```sh
aws dynamodb scan --table-name LaunchContext --endpoint-url http://localhost:8000
```
- [x] expires_on should be `now + expires_on` being reasonable close to what ever the current time in addition to the expires_on field should be a good indication.

## Refresh

**Test Steps**

- Go through refresh flow on default issuer.

**Expected Results**

```sh
aws dynamodb scan --table-name OAuthRequestsV2 --endpoint-url http://localhost:8000
```

should return following fields

- [x] code
- [x] internal_state
- [x] redirect_uri
- [x] state
- [x] expires_on | 42 days
- [x] access_token
- [x] refresh_token
- [x] iss

## Refresh Pull Document from OAuthRequestsV1

**Test Steps**

- Go through Token Flow
- Delete Token Item in OAuthRequestsV2
- Inject Code, State and Refresh_Token into OAuthRequests
- Go through refresh token flow.

**Expected Results**

```sh
aws dynamodb scan --table-name OAuthRequestsV2 --endpoint-url http://localhost:8080
```

should return following fields

- ? does not store code. should it?
- [x] internal_state
- [x] redirect_uri
- [x] state
- [x] expires_on | 42 days
- [x] access_token
- [x] refresh_token
- [x] iss

### DIFF
--- a/oauth-proxy/oauthHandlers/tokenHandlerStrategyClasses/saveDocumentStrategies/saveDocumentLaunchStrategy.js
+++ b/oauth-proxy/oauthHandlers/tokenHandlerStrategyClasses/saveDocumentStrategies/saveDocumentLaunchStrategy.js
@@ -1,4 +1,3 @@
-const jwtDecode = require("jwt-decode");
 const { hashString } = require("../../../utils");
 
 class SaveDocumentLaunchStrategy {
@@ -16,11 +15,10 @@ class SaveDocumentLaunchStrategy {
           this.config.hmac_secret
         );
 
-        let decodedToken = jwtDecode(tokens.access_token);
         let payload = {
           access_token: accessToken,
           launch: launch,
-          expires_on: decodedToken.exp,
+          expires_on: Math.round(Date.now() / 1000) + tokens.expires_in,
         };
 
         await this.dynamoClient.savePayloadToDynamo(
--- a/oauth-proxy/oauthHandlers/tokenHandlerStrategyClasses/saveDocumentStrategies/saveDocumentStateStrategy.js
+++ b/oauth-proxy/oauthHandlers/tokenHandlerStrategyClasses/saveDocumentStrategies/saveDocumentStateStrategy.js
@@ -1,5 +1,4 @@
 const { hashString } = require("../../../utils");
-const jwtDecode = require("jwt-decode");
 const { v4: uuidv4 } = require("uuid");
 
 class SaveDocumentStateStrategy {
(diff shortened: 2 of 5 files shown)


bdgrider-va (approved):
All unit and regression tests pass. Looks good.

beaugrantham-va (approved):

PR #267, opened 2021-03-25
upgrade openid-client from 3.15.10 to 4.4.2

esubocz-va:
Related story: https://vajira.max.gov/browse/API-5673
Upgrading openid-client dependency from version 3.15.10 to 4.4.2. 
Confirmed update with `npm outdated` and tested with regression script.

### DIFF
--- a/oauth-proxy/package-lock.json
+++ b/oauth-proxy/package-lock.json
@@ -22,7 +22,7 @@
         "jwt-decode": "^2.2.0",
         "morgan": "^1.10.0",
         "node-cache": "^5.1.2",
-        "openid-client": "^3.15.10",
+        "openid-client": "^4.4.2",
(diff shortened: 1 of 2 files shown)


crolarlibertyva (approved):

beaugrantham-va:
List of breaking-changes:
* https://github.com/panva/node-openid-client/releases/tag/v4.0.0

I believe the only one that may apply to us is:
> upgraded got http request library dependency from v9.x to v11.x. If you override some of the http request options you will most certainly have to accomodate them.

Can you verify this doesn't affect our custom options like timeout?

bdgrider-va (approved):
Verified npm outdated does not return open-id client, all tests passing.

beaugrantham-va (approved):

PR #275, opened 2021-04-02
API-5862 Add support for git-secrets

bdgrider-va:
Added initialization script, patterns file, allowed file, and readme description.


oauth-proxy/package-lock.json
```diff
@@ -1,9486 +1,8 @@
 {
   "name": "oauth-proxy",
   "version": "1.0.0",
-  "lockfileVersion": 2,
```
beaugrantham-va, line 4: Can you revert this file? It should not be affected by this story.

beaugrantham-va (changes requested):

.gitallowed
```diff
@@ -0,0 +1,8 @@
+.gitallowed
```
beaugrantham-va, line 1: Not sure why these two entries are in here?

bdgrider-va: Without these entries, the regex patterns would be picked up as false positive matches when running a scan.
image

beaugrantham-va (approved):

PR #279, opened 2021-04-08
API-5682 Remove unused acorn dependency.

bdgrider-va:
Removing acorn, a dependency that was unused. Verified all functionality with unit and regression tests.

https://vajira.max.gov/browse/API-5682

### DIFF
--- a/saml-proxy/package-lock.json
+++ b/saml-proxy/package-lock.json
@@ -1669,9 +1669,9 @@
       }
     },
     "acorn": {
-      "version": "7.4.0",
-      "resolved": "https://registry.npmjs.org/acorn/-/acorn-7.4.0.tgz",
-      "integrity": "sha512-+G7P8jJmCHr+S+cLfQxygbWhXy+8YTVGzAkpEbcLo2mLoL7tij/VG41QSHACSf5QgYRhMZYHuNc6drJaO0Da+w=="
+      "version": "8.1.0",
+      "resolved": "https://registry.npmjs.org/acorn/-/acorn-8.1.0.tgz",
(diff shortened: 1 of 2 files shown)


beaugrantham-va:
Breaking changes for 8.0.0

> ```
> ### Breaking changes
> 
> The `ecmaVersion` option is now required. For the moment, omitting it will still work with a warning, but that will change in a future release.
> 
> Some changes to method signatures that may be used by plugins.
> ```

https://github.com/acornjs/acorn/releases/tag/8.0.0

lukemarkwordtlibertyits (commented):
Does not appear to be used and all the regression and penetration testing passed. There does seem to still be references to acorn in the package.lock. Some dependencies use it as a sub-dependency. It is unclear whether these dependencies are used. For example `acorn-node` does not appear to be used. Maybe we could get rid of some of these dependencies aswell?

bdgrider-va:
> Does not appear to be used and all the regression and penetration testing passed. There does seem to still be references to acorn in the package.lock. Some dependencies use it as a sub-dependency. It is unclear whether these dependencies are used. For example `acorn-node` does not appear to be used. Maybe we could get rid of some of these dependencies aswell?

I see acorn-node as a child dependency of uswds, which we do need for styling. If this existed at the top level I would be all for removing it, but that is not the case.

beaugrantham-va (changes requested):
Please update the PR description to reflect that the dependency was removed (not updated).

lukemarkwordtlibertyits (approved):
Looks good

esubocz-va (approved):

beaugrantham-va (approved):

PR #282, opened 2021-04-12
API-5680 Remove node-sass

bdgrider-va:
Removed node-sass completely. This was not being used in our code. The node-sass-middleware package has its own child dependency of node-sass and there is no need for node-sass to exist at the top level.

Ran all linting and tests and confirmed expected functionality.

Attmpted to update the version of uswds, but it is a peer dependency of @department-of-veterans-affairs/formation and npm throws an error when I update it (even if updating formation to the latest version).

https://vajira.max.gov/browse/API-5680?focusedCommentId=881606&page=com.atlassian.jira.plugin.system.issuetabpanels:comment-tabpanel#comment-881606

### DIFF
--- a/saml-proxy/package-lock.json
+++ b/saml-proxy/package-lock.json
@@ -2897,6 +2897,11 @@
         }
       }
     },
+    "chownr": {
+      "version": "2.0.0",
+      "resolved": "https://registry.npmjs.org/chownr/-/chownr-2.0.0.tgz",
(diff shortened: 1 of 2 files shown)


lukemarkwordtlibertyits (approved):
Does not appear to be used. Both functional and visual regression tests passed.

esubocz-va (approved):

PR #283, opened 2021-04-12
Updated typescript dependencies

lukemarkwordtlibertyits:
# Description

Update saml-proxy dependencies related to typescript

- ts-node -> 9.1.1
- typescript -> 4.2.4
- @types/node -> 14.14.37

# Potential Breaking Changes

## ts-node
- 8 -> [9](https://github.com/TypeStrong/ts-node/releases/tag/v9.0.0)
  - Dropped support for Node 10
  - Re-add realpath, which should fix workflows that use symlinks, such as pnpm, Lerna workspaces, or other mono-repo configurations

## typescript
- 3 -> [4](https://devblogs.microsoft.com/typescript/announcing-typescript-4-0-beta/#breaking-changes)
  - Properties Overridding Accessors (and vice versa) is an Error
  - Operands for delete must be optional.
  - Usage of TypeScript’s Node Factory is Deprecated

## @types/node

- Also did not have a whole lot to work with for breaking changes.

## Suspicions about breaking changes 

I believe the version of node and typescript we are running will be important to these changes.

**What version of Node and Typescript are we running?**

- with these changes we will be running ts-node version 9.1.1 locally.
- We will be running typescript 4.2.4 in docker.
- We will be running node 12 in the dockerized version so it is probably important to support node 12 and up.

# Tests

These dependencies are used to run the saml-proxy. ts-node is used when runnin `npm run start-dev` and typescript is used when running in docker. The regression tests must pass in both type of runs

- [x] regression tests pass in local run
- [x] regression tests pass in dockerized run.


beaugrantham-va (dismissed):
Looks fine.

https://github.com/department-of-veterans-affairs/vets-saml-proxy/pull/285 and https://github.com/department-of-veterans-affairs/vets-saml-proxy/pull/284 should go first.

bdgrider-va (approved):
Looks good, all linting, unit tests, and regression tests pass.

crolarlibertyva (approved):

PR #284, opened 2021-04-14
Updating y18n dependency where possible.

bdgrider-va:
Followed recommended audit update command
`npm update y18n --depth 9`

Verified changes were non-breaking with linting, unit tests, and manual regression tests.
Using npm 6.12.14


crolarlibertyva:
So far so good. Confirmed just versions 4.0.3 and 5.0.8 appear.

beaugrantham-va:
This will need to be compatible with npm 6.14.12 until the project is run/built in a newer base container.

bdgrider-va:
> Might be worth updating the uswds dependency as well, because there is no story for that and it is the only dependency that upgrading will eliminate the older version of y18n

Looks like uswds has a child dependency that is no longer vulnerable. Going to leave it at this, with just the y18n fix, just to keep our changes small and easily revertible.

image

lukemarkwordtlibertyits (dismissed):
Looks good!

beaugrantham-va:
Looks fine, I'd like to get a functional sign-off from @lukemarkwordtlibertyits before merging.

lukemarkwordtlibertyits (approved):
Functionally looks good

beaugrantham-va (approved):

PR #291, opened 2021-04-22
API-5918 Remove unused css from saml-proxy

bdgrider-va:
Our sass middleware, **node-sass-middleware**, takes the .scss from the /styles directory and compiles it to .css in the /public directory. It also sets a prefix by which the css is addressable (/samlproxy/idp). The only .css file being requested (verified in grafana and local testing) is /samlproxy/idp/core.css. All others can be safely removed.

image


lukemarkwordtlibertyits (approved):
Things I considered when approving this PR.

- Full Regression and Penetration testing on this Branch
- Visual Regression Tests Ran. These test ensure the pixel diff between the template pages and the new pages do not exceed a certain threshold. For most of the error pages the threshold is relatively high at .7, but this is to account for the changing error ids.
  - Error diff does not exceed .7
  - Sensitive error diff does not exceed .7
  - Matching error diff does not exceed .7
  - IDP Selection diff does not exceed 0

PR #297, opened 2021-05-04
Update saml proxy UUID

lukemarkwordtlibertyits:
UUID dependency should be updated: 3.4.0 -> 8.3.2

# AC

- [x] unit tests pass
- [x] Integration tests pass
- [x] visual regression tests pass
- [x] requests should have unique IDs

### DIFF
--- a/saml-proxy/src/routes/index.js
+++ b/saml-proxy/src/routes/index.js
@@ -7,7 +7,7 @@ import cookieParser from "cookie-parser";
 import flash from "connect-flash";
 import sassMiddleware from "node-sass-middleware";
 import tildeImporter from "node-sass-tilde-importer";
-import uuidv4 from "uuid";
+import { v4 as uuidv4 } from "uuid";
 import rTracer from "cls-rtracer";
 import { RedisCache } from "./types";
 import {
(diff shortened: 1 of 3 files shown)


esubocz-va (dismissed):

saml-proxy/src/routes/index.js
```diff
@@ -7,7 +7,7 @@ import cookieParser from "cookie-parser";
 import flash from "connect-flash";
 import sassMiddleware from "node-sass-middleware";
 import tildeImporter from "node-sass-tilde-importer";
-import uuidv4 from "uuid/v4";
+import uuidv4 from "uuid";
```
beaugrantham-va, line 10: I don't believe this is correct. When I expirement, `uuidv4` is undefined.

According to the [docs](https://www.npmjs.com/package/uuid), this should be `import { v4 as uuidv4 } from 'uuid';`.

beaugrantham-va (changes requested):

saml-proxy/src/routes/index.js
lukemarkwordtlibertyits: I seemed to have confused uuid with rtracer. I will correct this.

lukemarkwordtlibertyits: So I found that with the incorrect import a session id will still be created but it will be a bit of a different format

**incorrect import**

image

**correct import**

image

lukemarkwordtlibertyits: I guess the express-session defaults to underscores; whereas, uuid.v4 does dashes

beaugrantham-va (approved):

PR #303, opened 2021-05-10
API-5707 switch to use auth0/passport-wsfed-saml2

crolarlibertyva:
Update to use original upstream project at https://github.com/auth0/passport-wsfed-saml2 instead of a fork at https://github.com/department-of-veterans-affairs/lighthouse-passport-wsfed-saml2

### DIFF
--- a/saml-proxy/test/testServer.js
+++ b/saml-proxy/test/testServer.js
@@ -5,7 +5,6 @@ import IDPConfig from "../src/IDPConfig";
 import configureExpress from "../src/routes";
 import { TestCache } from "../src/routes/types";
 import SPConfig from "../src/SPConfig";
-import { removeHeaders } from "../src/utils";
 
 import { idpCert, idpKey, spCert, spKey } from "./testCerts";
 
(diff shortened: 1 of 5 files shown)


saml-proxy/test/utils.test.js
```diff
@@ -0,0 +1,39 @@
+require("jest");
+
+import { getReqUrl, removeHeaders } from "../src/utils";
+import { idpCert } from "./testCerts";
+
+describe("Tests for utils.js", () => {
+  test("Test for removeHeaders", () => {
```
lukemarkwordtlibertyits, line 7: were these test added to increase code coverage, or is there something you discovered about these functions that you felt needed to be tested?

saml-proxy/test/testServer.js
```diff
@@ -23,7 +22,7 @@ const defaultTestingConfig = {
   spIdpIssuer: "api.idmelabs.com",
   spAuthnContextClassRef: "http://idmanagement.gov/ns/assurance/loa/3",
   spAcsUrls: ["/samlproxy/sp/saml/sso"],
-  idpCert: removeHeaders(idpCert),
```
lukemarkwordtlibertyits, line 26: why would switching to the non-forked passport library change this?

crolarlibertyva: The x509 data in the key section of the generated saml response was always null in the e2etests. In the e2etests, the forked library didn't seem to really verify the signatures on the signed assertions. So that test continued to pass. The non-forked library did attempted that check and error-ed when parsing out the cert data. And that caused the test failures.

lukemarkwordtlibertyits (approved):
All regression tests passed

PR #308, opened 2021-05-11
Add request ids to IDP saml requests

lukemarkwordtlibertyits:
For the purposes of trouble shooting, saml requests to IDPs should have unique RequestIDs

# Things to know

I made the request IDs match the express request ID. I think it would be helpful to be able to connect the messages to their express request.

# AC

- [x] Add e2e test
- [x] Verify requests IDs have been added to the saml requests
image


crolarlibertyva (approved):
Confirmed the presence of the ID field with BURP.

PR #311, opened 2021-05-12
API-6987 remove underscore dependency

crolarlibertyva:
Get rid if dependency on jsdoc.


crolarlibertyva:
Still a change to get in.

lukemarkwordtlibertyits (approved):
Regression tests are passing. JSDoc is running fine.

PR #312, opened 2021-05-13
API-7029 Forced resolution required for postcss

crolarlibertyva:
- Note that issue was the result of a deep dependency chain.
- There is a followup to this effort.
  - See https://vajira.max.gov/browse/API-7074

### DIFF
--- a/saml-proxy/package-lock.json
+++ b/saml-proxy/package-lock.json
@@ -2213,6 +2213,18 @@
         "num2fraction": "^1.2.2",
         "postcss": "^7.0.32",
         "postcss-value-parser": "^4.1.0"
+      },
+      "dependencies": {
+        "postcss": {
+          "version": "8.2.15",
+          "resolved": "https://registry.npmjs.org/postcss/-/postcss-8.2.15.tgz",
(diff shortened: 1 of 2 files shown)


lukemarkwordtlibertyits (commented):
I see the force resolution working in some places, but there still seems to be some versions less than 8.2.10 in the package-lock file. Is this an issue?

crolarlibertyva:
> I see the force resolution working in some places, but there still seems to be some versions less than 8.2.10 in the package-lock file. Is this an issue?

Let me double check. Wish npm audit didn't error out on me.

lukemarkwordtlibertyits:
> > I see the force resolution working in some places, but there still seems to be some versions less than 8.2.10 in the package-lock file. Is this an issue?
> 
> Confirmed that selected versions in the package-lock.json are 8.2.10
> 
> ```
>         "postcss": {
>           "version": "8.2.15",
> ```

I see where I was wrong in reading the package-lock.json file. Dependencies will list there respective versions, but they will actually be using the forced version.

lukemarkwordtlibertyits (approved):

PR #325, opened 2021-06-16
updated to uswds 1.6.14

lukemarkwordtlibertyits:
- Updated to uswds 1.6.14
- removed forced trim and postcss dependencies

# AC

- [x] regression tests passed

Note: There was a slight difference in the IDP selection page, but it may be due to me being on a new machine. The changes are only slightly in position, but is unnoticable to the user. See the heatmap below:
error_diff

### DIFF
--- a/saml-proxy/package-lock.json
+++ b/saml-proxy/package-lock.json
@@ -1,8 +1,13742 @@
 {
   "name": "saml-proxy",
   "version": "1.1.0",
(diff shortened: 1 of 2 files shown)


crolarlibertyva (approved):

dbrown-va (approved):
Successfully exercised locally with several test users.

PR #326, opened 2021-06-16
API-7819 update trim-newlines

crolarlibertyva:
- trim-newlines updated via resolutions
- note that the upstream dependency 'node-sass-middleware' which has an indirect dependency on 'trim-newlines' does not have a release available that addresses this desired upgrade


lukemarkwordtlibertyits (approved):
Confirmed functionality has not changed.

PR #327, opened 2021-06-17
API-8085 normalize-url >= 4.5.1

crolarlibertyva:
- Upgraded 'normalize-url' using resolutions.
- This was a transitive dependency via 'nodemon', of which this project is using the latest release

### DIFF
--- a/saml-proxy/package-lock.json
+++ b/saml-proxy/package-lock.json
@@ -84,7 +84,7 @@
         }
       }
     },
-    "@babel/generator": {å
+    "@babel/generator": {
       "version": "7.13.9",
       "resolved": "https://registry.npmjs.org/@babel/generator/-/generator-7.13.9.tgz",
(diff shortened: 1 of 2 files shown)


saml-proxy/package-lock.json
```diff
@@ -2884,6 +2884,12 @@
           "resolved": "https://registry.npmjs.org/lowercase-keys/-/lowercase-keys-2.0.0.tgz",
           "integrity": "sha512-tqNXrS78oMOE73NMxK4EMLQsQowWf8jKooH9g7xPavRT706R6bkQJ6DY2Te7QukaZsulxa30wQ7bk0pm4XiHmA==",
           "dev": true
+        },
+        "normalize-url": {
+          "version": "6.0.1",
```
beaugrantham-va, line 2889: It worries me a bit, bumping a transitive dependency 2 major versions. There is a risk of unknown breaking changes.

Version 4.5.1 is patched and part of the same major version. Should we use this instead?

beaugrantham-va (approved):
Have we got a process in place to review these forced resolutions for eventual removal?

crolarlibertyva:
> Have we got a process in place to review these forced resolutions for eventual removal?

As we upgrade direct dependencies, we always cross check to see if it allows us to remove an existing forced resolution.

PR #330, opened 2021-06-22
Api 7534 saml proxy reorg

crolarlibertyva:
lighthouse-saml-proxy reorganized as a distinct project in its own git repo


beaugrantham-va:
Are there other CICD changes that need to happen outside of this? I recall some work that had to be done for oauth-proxy. If so, can you link the stories/PRs?

saml-proxy/test/test-private-key.pem
```diff
@@ -1,27 +0,0 @@
------BEGIN RSA PRIVATE KEY-----
```
beaugrantham-va, line 1: Are these certs not needed for the unit or regression tests?

crolarlibertyva: yes they are.

crolarlibertyva:
> Are there other CICD changes that need to happen outside of this? I recall some work that had to be done for oauth-proxy. If so, can you link the stories/PRs?

 - https://github.com/department-of-veterans-affairs/devops/pull/9452

Tom temporarily applied the change so I can test out cicd for the repo.

lukemarkwordtlibertyits (approved):
LGTM

PR #333, opened 2021-07-08
API-7785 - Node 16

beaugrantham-va:
https://vajira.max.gov/browse/API-7785

- Updates Dockerfiles to use node 16 base image.
- Replaces node-sass and node-sass-middleware with dart-sass and custom middleware.
  - Note that the imported CSS from USWDS is using deprecated operations, resulting in warnings from dart-sass when rendering. Safe to ignore, for now
  - Upstream issue - https://github.com/uswds/uswds/issues/4204
  - Otherwise the generated CSS is functionally equivalent to the previous output.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -113,7 +113,6 @@ $RECYCLE.BIN/
 *.lnk
 
 ### Project ###
-
 idp*.pem
 sp*.pem
 okta*.pem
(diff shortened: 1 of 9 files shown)


crolarlibertyva (approved):
Wow!

lukemarkwordtlibertyits (changes requested):
Have you been able to do a fresh install locally with Node 16?

dbrown-va (approved):
Confirmed running locally and full regression tests.

lukemarkwordtlibertyits (approved):
LGTM

PR #334, opened 2021-07-15
Api 8349 publish regression tests

lukemarkwordtlibertyits:
# Description

Publish Saml-Proxy-Tests image to ECR.

## Note

- The node16 upgrade also upgraded the test image. This broke the test-image build. I didn't catch that, but the way we currently use puppeteer only works with centos 7 and below. The Node 14 and 16 images are based in centos 8. I had to downgrade for this to work.

### DIFF
--- a/Dockerfile.test
+++ b/Dockerfile.test
@@ -1,3 +1,4 @@
+# For now any node image that used CentOS > 7 does not work with puppeteer. We will have to keep this image at Node12.
 FROM vasdvp/lighthouse-node-application-base:node12
 
 # Build Args
--- a/Dockerfile.test
+++ b/Dockerfile.test
@@ -1,4 +1,4 @@
-FROM vasdvp/lighthouse-node-application-base:node16
+FROM vasdvp/lighthouse-node-application-base:node12
 
 # Build Args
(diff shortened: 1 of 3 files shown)


lukemarkwordtlibertyits:
I have pulled the ECR image locally and verified that it is working.

Dockerfile.test
```diff
@@ -1,4 +1,26 @@
-FROM vasdvp/lighthouse-node-application-base:node16
+# For now any node image that used CentOS > 7 does not work with puppeteer. We will have to keep this image at Node12.
+FROM vasdvp/lighthouse-node-application-base:node12
```
crolarlibertyva, line 2: Curious of the downgrade. Maybe node14 instead?

lukemarkwordtlibertyits: All bots use node12 atm, the reason is that both node14 and node16 use centos8 which for some reason does not support a dockerized browser. Definitely something that should be looked into, but the original container had node12, which got upgraded during the node16 upgrade. I did not catch that, but it broke the docker build. 

I will create Spike to investigate how to support a dockerized browser with centos8.

crolarlibertyva (approved):

PR #336, opened 2021-07-20
API-8350 refactor only

crolarlibertyva:
- This just refactors the method
- There will be a followup PR that includes the jest upgrade and the jest test changes

### DIFF
--- a/src/routes/handlers.js
+++ b/src/routes/handlers.js
@@ -59,6 +59,7 @@ export const samlLogin = function (template) {
       req.sp.options.getResponseParams(),
       new SAML.SAML(req.sp.options.getResponseParams())
     );
+    const authOptions = {};
     [
       ["id_me_login_link", "http://idmanagement.gov/ns/assurance/loa/3"],
       ["dslogon_login_link", "dslogon"],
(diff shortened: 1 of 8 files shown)


src/routes/handlers.js
```diff
@@ -173,3 +158,58 @@ export const handleError = (req, res) => {
   logger.error({ idp_sid: req.cookies.idp_sid });
   res.render(urlUserErrorTemplate(req), { request_id: rTracer.id() });
 };
+
+const getSamlRequestUrl = (options, params, exParams) => {
```
beaugrantham-va, line 162: This needs to be documented. I have no idea what sort of options, params, or exParams I can or must provide.

crolarlibertyva: Will do the same for other broken out functions.

src/routes/handlers.js
```diff
@@ -173,3 +158,58 @@ export const handleError = (req, res) => {
...
+          .then((url) => {
+            if (url) {
+              authOpts[key] = url;
+              resolve(authOpts);
+            }
+          })
+          .catch((err) => reject(err));
+      });
+    })
+    .catch(next);
```
beaugrantham-va, line 214: I'm not sure this error handling is functionally equivalent. It's hard to follow the chaining, but it looks like if a promise is rejected that this will hide it from the caller... who is also trying to `catch(next)`?

I'm not comfortable with these changes merging without accompanying unit level tests which cover error handling scenarios.

PR #341, opened 2021-07-27
Updated test docker file to node16

lukemarkwordtlibertyits:
# Description

Part of API-8698. Upgrade all images that use puppeteer to Node16.

- Saml-Proxy-Regression tests <- this PR
  - https://github.com/department-of-veterans-affairs/lighthouse-saml-proxy/pull/341
- Lighthouse-Auth-Utils 
  - https://github.com/department-of-veterans-affairs/lighthouse-auth-utils/pull/23

Note: The oauth-proxy regression tests use bats to test, so it is not included in this PR.

## To Ensure Proper Functionality

Run Regression Test Docker Image

image

### DIFF
--- a/Dockerfile.test
+++ b/Dockerfile.test
@@ -1,4 +1,3 @@
-# For now any node image that used CentOS > 7 does not work with puppeteer. We will have to keep this image at Node12.
 FROM vasdvp/lighthouse-node-application-base:node16
 
 # Build Args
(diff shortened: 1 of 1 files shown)


Dockerfile.test
```diff
@@ -1,5 +1,5 @@
 # For now any node image that used CentOS > 7 does not work with puppeteer. We will have to keep this image at Node12.
```
dbrown-va, line 1: We can remove this comment now that we are upgrading to CentOS8/Node16

dbrown-va (dismissed):
Confirmed both local `npm test` and docker run of the regression tests.

Dockerfile.test
lukemarkwordtlibertyits: Good catch

dbrown-va (approved):

PR #343, opened 2021-08-30
Check that image with tag exists prior to deployment

duganth-va:
* https://vajira.max.gov/browse/API-6867
* Adds small script to Makefile to check if tag exists in ECR
* Adds make target during deployment process to verify tag exists to prevent long lived deployment failures. 

Deployment with Tag that does not exist:https:
https://lighthouseva.slack.com/archives/C01CP09F04Q/p1630355720004200


dbrown-va (approved):
Confirmed clear error in the logs from the test deployment.

Looks good.

PR #347, opened 2021-09-24
API-9170 | Handle invalid assertion responses.

lukemarkwordtlibertyits:
# Description

The saml-proxy use to redirect to the `/login` page when an assertion response was invalid. The `login` functionality was not supported by the saml-proxy so it has since been deprecated. Now invalid assertion responses will return an error.

## Quick Start

```sh
curl http://localhost:7000/samlproxy/sp/saml/sso
```

if in web browser the response will look like this:

image

## TODO

- [x] functionality
- [x] Unit test change (or addition)
- [x] regression test change (or addition)


dbrown-va (approved):
Looks good.

Confirmed locally, acsHandlers.tests additions are thorough.

PR #348, opened 2021-09-29
API-8331 | add husky hooks

lukemarkwordtlibertyits:
# Description

Husky will be installed on`npm run prepare`, which run on `npm install`. Then the pre-commit script will check for aws-secret patterns and our local `/git-secrets-patterns`. If they are not it will add the values. Then it will run all the git-secrets scripts on commit. There is no way to run the oauth-proxy without initializing git secrets. For testing this branch you may want to run `npm run prepare` to avoid a fresh install.

## Potential future additions

It may be work running eslint on git pushes.

### DIFF
--- a/.husky/commit-msg
+++ b/.husky/commit-msg
@@ -1,0 +1,4 @@
+#!/bin/sh
+. "$(dirname "$0")/_/husky.sh"
+
+git secrets --commit_msg_hook -- "$@"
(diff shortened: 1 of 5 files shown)


lukemarkwordtlibertyits:
I cloned a fresh repo and attempted to check in a file that broke the secrets-pattern

image

dbrown-va (dismissed):
Confirmed similar behavior locally following an `npm install`.  Pretty neat.

image

dbrown-va (approved):

PR #351, opened 2021-10-14
API-10490 login gov link UI request

crolarlibertyva:
* Use dev-config.json sent in keybase to include login_gov entry
  * login page should include login_gov entry
  * when selecting login_gov entry will be sent to login_gov site to log in
* Use original config to deploy w/o login_gov entry
  * login page should render w/o login_gov entry 
 
PR in draft, additional unit tests pending.


src/routes/routes.js
```diff
@@ -46,15 +46,19 @@ export default function addRoutes(
   });
 
   app.get(SP_METADATA_URL, function (req, res, next) {
-    const xml = METADATA_TEMPLATE(spConfig.getMetadataParams(req));
+    // Other metadata options?
```
beaugrantham-va, line 49: Is this comment needed?

crolarlibertyva: Will take it out.

src/routes/routes.js
```diff
@@ -46,15 +46,18 @@ export default function addRoutes(
...
+    const xml = METADATA_TEMPLATE(spConfigs.id_me.getMetadataParams(req));
     res.set("Content-Type", "text/xml");
     res.send(xml);
   });
 
   app.get(SP_VERIFY, parseSamlRequest, samlLogin("verify"));
 
-  spConfig.acsUrls.forEach((url) => acsFactory(app, url, cache, cacheEnabled));
-
+  Object.values(spConfigs).forEach((spConfig) => {
```
beaugrantham-va, line 56: I don't understand what this is doing.

There is currently only one ACS URL so it seems like this is not doing anything more than the line that was removed above.

src/app.js
```diff
@@ -53,12 +58,27 @@ const handleMetadata = (argv) => {
...
   IdPMetadata.fetch(argv.spIdpMetaUrl)
     .then(handleMetadata(argv))
     .then(() => {
       const app = express();
       const httpServer = http.createServer(app);
-      const spConfig = new SPConfig(argv);
+      const spConfigs = { id_me: new SPConfig(argv) };
+      strategies.id_me = createPassportStrategy(spConfigs.id_me);
+      app.use(passport.initialize());
+      if (argv.otherLogins) {
```
lukemarkwordtlibertyits, line 70: Changes in dev-config.base.json and src/cli/index.js should be made to indicate this added config field.

crolarlibertyva: Roger that.

src/routes/routes.js
lukemarkwordtlibertyits: I agree that even the original method is a little strange ... there should only be one ACS url right?

dbrown-va:
I am getting an Okta error when attempting to login with ID.me

In Okta, seeing `Unable to validate incoming SAML Assertion`.

dev-config.base.json
```diff
@@ -23,5 +23,18 @@
...
   "redisHost": "127.0.0.1",
-  "spIdpSsoBinding": "urn:oasis:names:tc:SAML:2.0:bindings:HTTP-Redirect"
+  "spIdpSsoBinding": "urn:oasis:names:tc:SAML:2.0:bindings:HTTP-Redirect",
+  "otherLogins": 
+  [
+    {
+      "category": "login_gov",
+      "spIdpMetaUrl": "https://idp.int.identitysandbox.gov/api/saml/metadata2021",
+      "spNameIDFormat": "urn:oasis:names:tc:SAML:2.0:nameid-format:persistent",
+      "spAudience": "urn:gov:gsa:SAML:2.0.profiles:sp:sso:va_lighthouse:saml_proxy_local",
```
beaugrantham-va, line 33: This base config should contain generic entries.

src/cli/index.js
```diff
@@ -236,11 +236,11 @@ export function processArgs() {
         string: true,
         default: "Simple SAML Service Provider",
       },
-      spAcsUrls: {
-        description: "SP Assertion Consumer Service (ACS) URLs (Relative URL)",
+      spAcsUrl: {
```
beaugrantham-va, line 239: This needs to be updated in the sample base config.

dev-config.base.json
```diff
@@ -23,5 +23,18 @@
...
+      "spAuthnContextClassRef": "http://idmanagement.gov/ns/assurance/loa/3",
+      "spCert": ".saml-proxy-public-local-login-gov.crt",
+      "spKey": "./saml-proxy-private-local-login-gov.key"
+   }
+  ]
 }
+
```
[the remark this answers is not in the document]
beaugrantham-va: The README also has a lot of content which assumes that ID.me is the only direct IDP/CSP. This needs to be updated at some point. I am okay with this being done under another PR but we need to track this.

src/routes/handlers.js
```diff
@@ -68,9 +78,21 @@ export const samlLogin = function (template) {
         "http://idmanagement.gov/ns/assurance/loa/3",
         "&op=signup",
       ],
-    ]
+    ];
+    if (login_gov_enabled) {
+      authnSelection.push([
+        "login_gov_login_link",
+        "http://idmanagement.gov/ns/assurance/loa/3",
```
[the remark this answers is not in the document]
dbrown-va: Agreed, this would be `http://idmanagement.gov/ns/assurance/ial/2` from the Login.gov [docs](https://developers.login.gov/saml/#identity-assurance-level-ial)

Looks like they default require 2FA.

> urn:gov:gsa:ac:classes:sp:PasswordProtectedTransport:duo 
> This specifies that a user has been authenticated with a second factor. This value will be returned in the user attributes by default. We do not allow AAL 1, because it implies that a user did not authenticate with a second factor.

dev-config.base.json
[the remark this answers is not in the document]
crolarlibertyva: https://vajira.max.gov/browse/API-10735

src/app.js
```diff
@@ -41,6 +43,9 @@ const handleMetadata = (argv) => {
           if (metadata.signRequest) {
             argv.spSignAuthnRequests = metadata.signRequest;
           }
+          if (metadata.NameIDFormat) {
```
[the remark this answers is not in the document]
crolarlibertyva: It's supposed to be nameIdFormats which is array. Have to redo this or take it out.

dev-config.base.json
```diff
@@ -23,5 +23,18 @@
   "cacheEnabled": true,
   "redisPort": "6379",
   "redisHost": "127.0.0.1",
-  "spIdpSsoBinding": "urn:oasis:names:tc:SAML:2.0:bindings:HTTP-Redirect"
+  "spIdpSsoBinding": "urn:oasis:names:tc:SAML:2.0:bindings:HTTP-Redirect",
+  "otherLogins":
```
dbrown-va, line 27: This should be updated to `idpSamlLogins`

src/routes/handlers.js
```diff
@@ -55,11 +55,21 @@ export const samlLogin = function (template) {
...
+        new SAML.SAML(req.sps.options[idpKey].getResponseParams())
+      );
+    });
+
+    let login_gov_enabled;
+    if (req.sps.options.login_gov) {
+      login_gov_enabled = true;
+    } else {
+      login_gov_enabled = false;
+    }
```
dbrown-va, line 71: Does this take into account the top level `argv.otherLoginsEnabled` value being set?

dbrown-va (changes requested):
Would like to get alignment on the flag name and the actual idpSaml config options.

Also added a question on the 2 different ways the login.gov csp is enabled/disabled.  Wanted to make sure that was working like we expect.

PR #356, opened 2021-10-27
Api-000 | refactor idp selection

lukemarkwordtlibertyits:
# Description

IDP category is injected in the claims, so we can determine what idp is being used.


src/IDPProfileMapper.ts
```diff
@@ -95,6 +102,18 @@ const logonGovConfiguration: IClaimDescriptions = {
     displayName: "Authentication Assurence Level",
     description: "Method in which user should be authenticated",
     multiValue: false,
+    transformer: (claims: { aal?: String }) => {
+      if (claims.aal && claims.aal === PASSWORDPROTOCOL.DEFAULT) {
```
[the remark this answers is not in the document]
lukemarkwordtlibertyits: Decided to go the route you suggested in the slack chat. We will determine AAL by string and return 0 if there is no match so it will not pass the LOA check.

src/IDPProfileMapper.ts
```diff
@@ -109,7 +129,7 @@ const logonGovConfiguration: IClaimDescriptions = {
           return parseInt(parsedIal);
         }
       }
-      return undefined;
+      return 0;
```
beaugrantham-va, line 132: I think we should do the same explicit matching here as we do with `aal`.

src/IDPProfileMapper.ts
```diff
@@ -119,8 +139,14 @@ const logonGovConfiguration: IClaimDescriptions = {
...
     transformer: (claims: { aal?: string }) => {
-      if (claims && claims.aal) {
-        if (SUFFICIENT_AAL.includes(claims.aal)) {
+      const multifactor = [
+        PASSWORDPROTOCOL.DEFAULT,
+        PASSWORDPROTOCOL.MULTIFACTOR_TWELVE,
+        PASSWORDPROTOCOL.CRYPTOGRAPHICALLYSECURE,
+        PASSWORDPROTOCOL.HSPD12,
+      ];
+      if (claims.aal && claims.aal) {
```
beaugrantham-va, line 148: Duplicative check. Can you verify that linting is actually being performed? I would expect this to be caught.

dbrown-va (commented):
Confirmed with Login.gov and ID.me tests including token validation.

Profile populates and is returned correctly.

src/IDPProfileMapper.ts
```diff
@@ -315,14 +341,23 @@ export class IDPProfileMapper implements ISamlpProfileMapper {
...
+    ) {
       this.getClaimFields(dsLogonConfiguration, claims);
-    } else if (this.samlAssertions.claims.ial) {
-      this.getClaimFields(logonGovConfiguration, claims);
-    } else {
+    } else if (this.samlAssertions.claims.category === "id_me") {
       this.getClaimFields(idmeConfiguration, claims);
+    } else if (
+      this.samlAssertions.claims.category === "logingov" &&
+      this.samlAssertions.claims.ial
```
beaugrantham-va, line 358: Why does this also check `ial`? Is this legacy code that should be removed after adding `category`?

src/IDPProfileMapper.ts
beaugrantham-va, line 132: The `displayName` has a typo.

src/IDPProfileMapper.ts
```diff
@@ -95,6 +102,19 @@ const logonGovConfiguration: IClaimDescriptions = {
     displayName: "Authentication Assurence Level",
     description: "Method in which user should be authenticated",
     multiValue: false,
+    transformer: (claims: { aal?: String }) => {
```
beaugrantham-va, line 105: The `displayName` has a typo.

beaugrantham-va (approved):

PR #358, opened 2021-11-04
API-000 refactor handler prep legacy config flag

crolarlibertyva:
- Add a new flag used to aid in the transition to a new config format
- Flag is defaulted to true
- It is required to be activated in the near term, pending subsequent refactoring

### DIFF
--- a/src/app.js
+++ b/src/app.js
@@ -61,8 +61,11 @@ function runServer(argv) {
     .then(() => {
       const app = express();
       const httpServer = http.createServer(app);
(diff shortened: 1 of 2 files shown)


dbrown-va (changes requested):
We would also want that value in the [base config](https://github.com/department-of-veterans-affairs/lighthouse-saml-proxy/blob/master/dev-config.base.json).

Also not sure if we confirmed going this direction just yet.  So may want to hold for now on this?

crolarlibertyva:
Will close until consensus on direction is reached.

PR #360, opened 2021-11-05
API-10491 login gov sign up

crolarlibertyva:
- Add a button on the saml-proxy UI for a user to register a login.gov account
  - This signup button can be enabled or disabled per environment
- In this case the signup page is the login page at login.gov. From there the user can press the "Create an account" button
Screen Shot 2021-11-09 at 2 34 43 PM

### DIFF
--- a/views/login_selection.hbs
+++ b/views/login_selection.hbs
@@ -23,7 +23,7 @@
     <div class="usa-width-one-half">
       <h4>Don't have these accounts?</h4>
       <a class="no-external-icon usa-button idme-signup" href="{{ id_me_signup_link }}"><img alt="ID.me" src="/samlproxy/idp/img/idme-icon-dark.svg" />Create an ID.me account</a>
-      {{#if login_gov_signup_link_enabled}}
+      {{#if login_gov_enabled}}
       <a class="no-external-icon usa-button logingov-signup" href="{{ login_gov_signup_link }}"><img alt="Login.gov" src="/samlproxy/idp/img/icon-dot-gov.svg" />Create a Login.gov account</a>
       {{/if}}
       <p>Use your email, Google, or Facebook</p>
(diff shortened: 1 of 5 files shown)


dbrown-va (changes requested):
Rather than directly navigating to the Login.gov create an account functionality, the Lighthouse `Create a Login.gov account` should still send the SAML request as if the user was attempting to login with that csp.

https://github.com/department-of-veterans-affairs/lighthouse-saml-proxy-configs/pull/21#pullrequestreview-800666506

views/login_selection.hbs
```diff
@@ -23,6 +23,9 @@
     <div class="usa-width-one-half">
       <h4>Don't have these accounts?</h4>
       <a class="no-external-icon usa-button idme-signup" href="{{ id_me_signup_link }}"><img alt="ID.me" src="/samlproxy/idp/img/idme-icon-dark.svg" />Create an ID.me account</a>
+      {{#if login_gov_signup_link_enabled}}
+      <a class="no-external-icon usa-button logingov-signup" href="{{ login_gov_signup_link }}"><img alt="Login.gov" src="/samlproxy/idp/img/icon-dot-gov.svg" />Create an login.gov account</a>
```
dbrown-va, line 27: `Create **a** login.gov account`

dbrown-va:
Can we have details for the story added to the description?

views/login_selection.hbs
```diff
@@ -23,6 +23,9 @@
     <div class="usa-width-one-half">
       <h4>Don't have these accounts?</h4>
       <a class="no-external-icon usa-button idme-signup" href="{{ id_me_signup_link }}"><img alt="ID.me" src="/samlproxy/idp/img/idme-icon-dark.svg" />Create an ID.me account</a>
+      {{#if login_gov_enabled}}
```
dbrown-va, line 26: VA.gov specifically shows/hides the sign up button separate from the Login.gov login itself.

Can we update so that button can be independently shown/hidden as well?

dbrown-va:
Confirmed button added to SAML Proxy only when enabled.  Successfully logged or initiated the create account flow on the Login.gov side.

image

dbrown-va (approved):
Confirmed behavior with button specifically enabled & hidden as well as login.gov overall.

Also confirmed the saml request for "Create an account" can successfully login or start the create account flow hosted on the Login.gov side.

PR #361, opened 2021-11-09
API-10489 | Create a refactored idp selection page

lukemarkwordtlibertyits:
# Description

See above.

## Quick start

Add `"idpSelectionRefactor" : true,` to the configuration.


dbrown-va:
Would like to see a bit more info in the description here.

A link to the VA.gov vets-website page to reference would be good.

crolarlibertyva (commented):
Looks just like the screenshot passed around in the thread to me. But it is different from what the va actually deployed to dev.

lukemarkwordtlibertyits:
> Looks just like the screenshot passed around in the thread to me. But it is different from what the va actually deployed to dev.

Derek and I agreed that we would wait until Va.gov had a final product to finalize the UI.

crolarlibertyva (approved):

views/login_selection_refactor.hbs
```diff
@@ -0,0 +1,109 @@
...
+                src="/samlproxy/idp/img/idme-icon-white.svg" />Sign in with ID.me</a>
+            <a class="no-external-icon usa-button dslogon" href="{{ login_gov_login_link }}"><img alt="Login.gov"
+                src="/samlproxy/idp/img/login-gov-logo.svg" />Sign in with Login.gov</a>
+            <div>
+              <span class="sidelines">OR</span>
+              <div class="alternate-signin">
+                <h2 class="vads-u-font-size--sm vads-u-margin-top--0">
+                  Don't have those accounts?
+                </h2>
+                <a class="no-external-icon usa-button signup" href="{{ login_gov_signup_link }}"><img alt="Login.gov" src="/samlproxy/idp/img/login-gov-dark-logo.svg" />Create a Login.gov account</a>
```
dbrown-va, line 36: It looks like this no longer takes into consideration the `spIdpSignupLinkEnabled` that was added in the previous PR.

lukemarkwordtlibertyits: So this is an entired refactored handlebars template, if the flag is false then the old template with that option is enabled. My thought was that since this template was the refactored template the `spIdpSignupLink` should always be enabled. If you didnt want it enabled you would use the old template. I can change this if desired.

dbrown-va: Got it.  We do want to maintain the ability to show/hide that Sign-Up independent of the refactor.  VA.gov will be rolling out the login option a couple weeks prior to the sign up.

Especially since if `spIdpSignupLinkEnabled` is set to false, the button ends up displayed but doesn't actually work.

dbrown-va (approved):
Confirmed logins with ID.me and Login.gov using the refactored and current pages.

Confirmed show/hide of `spIdpSignupLinkEnabled` and `idpSamlLoginsEnabled`.

crolarlibertyva (approved):

crolarlibertyva:
> Confirmed logins with ID.me and Login.gov using the refactored and current pages.
> 
> Confirmed show/hide of `spIdpSignupLinkEnabled` and `idpSamlLoginsEnabled`.

Same.

PR #367, opened 2021-11-23
API-11508 forced update to http-signature version

crolarlibertyva:
- Addresses a security update to json-schema via an upgrade to http-signature.
- Note, this was done via a forced resolution
- This is a short term fix
  - A longer term solution will be addressed by https://vajira.max.gov/browse/API-11511

- [x] Regression tests passed

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -8945,12 +8945,10 @@
       }
     },
     "node_modules/http-signature": {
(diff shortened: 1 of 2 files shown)


lukemarkwordtlibertyits (approved):
LGTM

PR #369, opened 2021-11-24
forced update for json-schema

crolarlibertyva:
- Addresses a security update to json-schema
- Note, this was done via a forced resolution
- This is a short term fix
  - A longer term solution will be addressed by https://vajira.max.gov/browse/API-11511

- [x] Regression tests passed

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -2286,9 +2286,7 @@
       }
     },
     "node_modules/jsprim": {
-      "version": "1.4.1",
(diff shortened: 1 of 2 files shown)


lukemarkwordtlibertyits (approved):
LGTM

dbrown-va (commented):
Successful login locally, no issues.  Follow up request story in backlog to address more permanent solution sounds good.

PR #372, opened 2022-01-14
API-11368 UI style minimal refactor

crolarlibertyva:
Minimal refactor with style updates for signin and signup selector
Screen Shot 2022-01-17 at 3 08 38 PM


dbrown-va (approved):
Confirmed existing behavior is unchanged and the minimal additional of Login.gov when the refactor is enabled.

Styling matches VA.gov to an extent, with limitation acknowledged due to the use of the modal in that login method.

PR #374, opened 2022-01-27
API-12695 force authn

crolarlibertyva:
- Temporary use of forceAuthn flag to enable the LoginGov IDP via the auth request
  - This is for login.gov testing in prod prior to activation
- This is a draft PR with an alternative approach in mind us


dbrown-va:
Recommend closing in favor of https://github.com/department-of-veterans-affairs/lighthouse-saml-proxy/pull/376.

This works approach would technically also work so good to have alternatives.

PR #376, opened 2022-01-28
API-12695 cookie activation

crolarlibertyva:
- Temporary use of a matching cookie to enable the LoginGov IDP via the auth request
  - This is for login.gov testing in prod prior to activation
- This is approach nominated for enabling the login.gov idp via the auth  request

Example config segment:
```
{
   .
   .
   .
   "idpSamlLoginsEnabled": true,
    "idpSelectionRefactor" : true,
    "idpSamlLogins": 
      [
        {
          "category": "logingov",
          "idpTemporarilyDisabled": true,
          "idpActivationCookie": "doublestuffedoreo",
          "spIdpMetaUrl": "https://idp.int.identitysandbox.gov/api/saml/metadata2021",
          "spNameIDFormat": "urn:oasis:names:tc:SAML:2.0:nameid-format:persistent",
          "spAudience": "urn:gov:gsa:SAML:2.0.profiles:sp:sso:va_lighthouse:saml_proxy_local",
          "spRequestAuthnContext": true,
          "spRequestNameIDFormat": true,
          "spCert": "./sp-cert.pem",
          "spKey": "./sp-key.pem",
          "spProtocol": "samlp",
          "spAuthnContextClassRef": "http://idmanagement.gov/ns/assurance/ial/2",
          "spIdpSignupLinkEnabled": true
        }
      ]
  }
```

### DIFF
--- a/src/routes/handlers.js
+++ b/src/routes/handlers.js
@@ -211,10 +211,7 @@ function enabled_logingov(req) {
     if (
       !req.sps.options.logingov.idpTemporarilyDisabled ||
       (req.sps.options.logingov.idpActivationCookie &&
-        req.cookies[req.sps.options.logingov.idpActivationCookie] &&
-        req.cookies[req.sps.options.logingov.idpActivationCookie].includes(
-          req.sps.options.logingov.category
-        ))
+        req.cookies[req.sps.options.logingov.idpActivationCookie])
     ) {
       login_gov_enabled = true;
     }
--- a/src/routes/handlers.js
+++ b/src/routes/handlers.js
@@ -209,7 +209,7 @@ function enabled_logingov(req) {
   req.cookies;
   if (req.sps.options.logingov) {
     if (
-      !req.sps.options.logingov.disabled ||
+      !req.sps.options.logingov.idpTemporarilyDisabled ||
       (req.sps.options.logingov.idpActivationCookie &&
         req.cookies[req.sps.options.logingov.idpActivationCookie] &&
         req.cookies[req.sps.options.logingov.idpActivationCookie].includes(
--- a/src/routes/handlers.js
+++ b/src/routes/handlers.js
@@ -64,7 +64,7 @@ export const samlLogin = function (template) {
     });
 
     let login_gov_enabled;
-    if (req.sps.options.logingov) {
+    if (req.sps.options.logingov && !req.sps.options.logingov.disabled) {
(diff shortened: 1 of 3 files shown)


src/routes/handlers.js
```diff
@@ -63,12 +63,7 @@ export const samlLogin = function (template) {
       );
     });
 
-    let login_gov_enabled;
-    if (req.sps.options.logingov) {
-      login_gov_enabled = true;
-    } else {
-      login_gov_enabled = false;
-    }
+    let login_gov_enabled = enabled_logingov(req, relayState);
```
dbrown-va, line 66: I think `relayState` snuck in here from some earlier dev.  Doesn't appear to be used in the method.

src/routes/handlers.js
```diff
@@ -208,3 +203,18 @@ export const handleError = (req, res) => {
   logger.error({ idp_sid: req.cookies.idp_sid });
   res.render(urlUserErrorTemplate(req), { request_id: rTracer.id() });
 };
+
+function enabled_logingov(req) {
+  let login_gov_enabled = false;
+  req.cookies;
```
dbrown-va, line 209: I think this may have been a debugging item or errant copy/paste.

esubocz-va:
failed build?

crolarlibertyva:
> failed build?
 - That's related to 
https://lighthouseva.slack.com/archives/C010ENX8X7G/p1643665422081739
- https://github.com/department-of-veterans-affairs/lighthouse-saml-proxy/pull/379 should resolve this.

dbrown-va:
This looks to work well as a temporary solution until we turn on login.gov as an IDP for everyone.

Could we get a corresponding PR for the configs repo that enables in Sandbox with this functionality?

If we could also get a unit test in as a sanity check, that would be great.

crolarlibertyva:
> This looks to work well as a temporary solution until we turn on login.gov as an IDP for everyone.
> 
> Could we get a corresponding PR for the configs repo that enables in Sandbox with this functionality?
> 
> If we could also get a unit test in as a sanity check, that would be great.

https://github.com/department-of-veterans-affairs/lighthouse-saml-proxy-configs/pull/26

dbrown-va (approved):
Works well.  Only renders login.gov option when cookie present.  Solves the problem of testing since it is the Authz server calling the saml proxy, not us directly.

PR #379, opened 2022-01-31
API-000 ci temp workaround fixed testimage version

crolarlibertyva:
- Address ci block from not being able to build the test image in ci


dbrown-va (approved):
Temporary fix until centos8 issue resolved

PR #384, opened 2022-02-25
Choose active signing key if more than 1

crolarlibertyva:
Choose the signing key marked as active if more than 1 is available in the IDP metadata

### DIFF
--- a/src/app.js
+++ b/src/app.js
@@ -24,10 +24,17 @@ const handleMetadata = (argv) => {
   return (metadata) => {
     if (metadata.protocol) {
       argv.spProtocol = metadata.protocol;
-      if (metadata.signingKeys[0]) {
-        argv.spIdpCert = cli.certToPEM(metadata.signingKeys[0]);
+      if (metadata.signingKeys) {
+        var signingKeyCert;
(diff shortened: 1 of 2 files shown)


dbrown-va (approved):
Confirmed with both CSP metadata options locally.

src/app.js
```diff
@@ -24,10 +24,17 @@ const handleMetadata = (argv) => {
   return (metadata) => {
     if (metadata.protocol) {
       argv.spProtocol = metadata.protocol;
-      if (metadata.signingKeys[0]) {
-        argv.spIdpCert = cli.certToPEM(metadata.signingKeys[0]);
```
beaugrantham-va, line 28: I don't understand how this worked before? It looks like `certToPEM` expects a string but this passed in an object?

src/app.js
```diff
@@ -24,10 +24,17 @@ const handleMetadata = (argv) => {
...
+      if (metadata.signingKeys) {
+        var signingKeyCert;
```
beaugrantham-va, line 28: This should be `let` or `const`  instead of `var`.

PR #385, opened 2022-02-25
API-13211 Single docker file for saml-proxy

crolarlibertyva:
Single docker file for saml-proxy to ensure parity between local testing with docker with the image that gets deployed


DockerfileFG.dockerignore
```diff
@@ -1,9 +0,0 @@
-public/bower_components/*
```
dbrown-va, line 1: Confirmed these components are all contained in the main `.dockeringore` [file](https://github.com/department-of-veterans-affairs/lighthouse-saml-proxy/blob/master/.dockerignore)

docker-compose.yml
```diff
@@ -14,4 +14,6 @@ services:
       dockerfile: Dockerfile
     ports:
       - '7000:7000'
-    command: 'node build/app.js --config dev-config.json'
+    volumes:
+      - ./dev-config.json:/etc/saml-proxy/config.json
```
dbrown-va, line 18: Confirmed this volume mounting working locally.

dbrown-va (approved):

PR #387, opened 2022-03-05
API-13140 Upgrade Node-Forge

johnmweisz:
Upgrade node-forge package at least v1.0.0 to remove security vulnerabilities.

Issue: https://vajira.max.gov/browse/API-13140

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -6170,9 +6170,7 @@
       }
     },
     "node_modules/node-forge": {
-      "version": "0.10.0",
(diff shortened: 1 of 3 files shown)


dbrown-va (changes requested):
The regression tests package.json will likely need a similar fix for node-forge.
https://github.com/department-of-veterans-affairs/lighthouse-saml-proxy/blob/master/test/regression_tests/package.json

crolarlibertyva (approved):

PR #388, opened 2022-03-23
API-14287 Update FAQ Link

johnmweisz:
Update FAQ link to point to production instead of staging.

issue: https://vajira.max.gov/browse/API-14287

### DIFF
--- a/views/login_selection.hbs
+++ b/views/login_selection.hbs
@@ -89,7 +89,7 @@
   </div>
   <div class="usa-width-one-half login-help">
     <h4>Having trouble signing in?</h4>
-    <p><a href="https://staging.va.gov/sign-in-faq/">Get answers to Frequently Asked Questions</a></p>
+    <p><a href="https://va.gov/sign-in-faq/">Get answers to Frequently Asked Questions</a></p>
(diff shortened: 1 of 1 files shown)


views/login_selection.hbs
```diff
@@ -89,7 +89,7 @@
   </div>
   <div class="usa-width-one-half login-help">
     <h4>Having trouble signing in?</h4>
-    <p><a href="https://staging.va.gov/sign-in-faq/">Get answers to Frequently Asked Questions</a></p>
+    <p><a href="https://va.gov/sign-in-faq/">Get answers to Frequently Asked Questions</a></p>
```
beaugrantham-va, line 92: This totally works. But, when you go to this link it actually follows a few redirects to a new location.

```
< HTTP/1.1 301 Moved Permanently
< Location: https://www.va.gov/sign-in-faq/

< HTTP/1.1 301 Moved Permanently
< Location: https://www.va.gov/resources/signing-in-to-vagov/
```

I would recommend updating the link to that last redirect in case VA.gov ever removes the old URL.

johnmweisz: noted, thanks.

johnmweisz:
Updated link to bypass redirect

crolarlibertyva (approved):

PR #390, opened 2022-03-29
API-13875 remove modify test

crolarlibertyva:
- Address https://vajira.max.gov/browse/API-13875 
- Removes the modify test in the the regression test suite
- Note that the removed test didn't exercised the ModifyAttack from saml-utils
     - Confirmed in debug runs


dbrown-va:
The modify test has been passing the last several regression test runs.  Recommend closing this PR for now.

PR #393, opened 2022-03-30
API-14655 Remove Login.gov Cookie

johnmweisz:
Removes idpActivationCookie and idpTemporarilyDisabled for Login.gov testing with cookie and related tests.

issue: https://vajira.max.gov/browse/API-14655

Follow up by removing idpTemporarilyDisabled from saml-proxy-configs


dbrown-va (approved):
Confirmed removal runs as expected locally.

In a future story, we should move the other idp's into that `idpSamlLogins` array so each can be shown / hidden in this fashion.

PR #394, opened 2022-05-19
API-16120 async upgrade

crolarlibertyva:
Addresses https://vajira.max.gov/browse/API-16120
 - Upgrade samlp and winston
   - These are the libs that require async
 - Upgrade async with forced resolution


dbrown-va (approved):
Confirmed `npm audit` no longer lists `async` as vulnerability locally.

Ran regression test locally passing outside of the replay test, also manually login working.

PR #397, opened 2022-07-13
moment >=2.29.4

johnmweisz:
fix vulnerability
https://vajira.max.gov/browse/API-17422


crolarlibertyva (commented):
Can you commit the package-lock.json that goes with this change?

crolarlibertyva (approved):

PR #398, opened 2022-07-14
API-17423 update got

crolarlibertyva:
Addresses https://vajira.max.gov/browse/API-17423
 - Updating package-lock using 'npm audit fix' removed the dependency on 'got' altogether.


package.json
```diff
@@ -134,6 +134,7 @@
     "typescript": "^4.2.4"
   },
   "resolutions": {
+    "got": ">=11.8.5",
```
johnmweisz, line 137: Does got require a forced resolution? I noticed that "npm audit fix" seems to fix it.

crolarlibertyva: Will check. I thought it was a cascade dependency.

crolarlibertyva: Good catch. The "audit fix" removed the cascading dependency on 'got'.

johnmweisz (approved):

PR #399, opened 2022-07-14
API-17423 passport upgrade

crolarlibertyva:
Addresses https://vajira.max.gov/browse/API-17423
 - Upgrades passport to >= 0.6.0

### DIFF
--- a/src/routes/acsHandlers.ts
+++ b/src/routes/acsHandlers.ts
@@ -71,8 +71,10 @@ export const buildPassportLoginHandler = (acsURL: string) => {
         state: req.query.RelayState || req.body.RelayState,
         url: getReqUrl(req, acsURL),
       };
-      if (req.session) {
-        req.session.ssoResponse = ssoResponse;
+      if (req.options) {
+        req.options.ssoResponse = ssoResponse;
+      } else {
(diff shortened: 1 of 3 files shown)


src/routes/acsHandlers.ts
```diff
@@ -71,8 +71,10 @@ export const buildPassportLoginHandler = (acsURL: string) => {
         state: req.query.RelayState || req.body.RelayState,
         url: getReqUrl(req, acsURL),
       };
-      if (req.session) {
-        req.session.ssoResponse = ssoResponse;
+      if (req.options) {
```
johnmweisz, line 74: `req.session ` is also found in 
- src/routes/index.js 
- src/routes/handlers.js 
- src/routes/acsHandlers.ts
Just wanted to confirm this is ok?

crolarlibertyva: Yeah, for some reason, the upgrade was causing the ssoResponse to get stripped out of the session. Since the ssoResponse a payload is passed from one stage to the next in the saml-proxy, I opted to stash it to a different object, "options", within the request. I only altered for the ssoResponse. Left the rest of the session object intact.

johnmweisz (approved):

PR #402, opened 2022-10-17
added error handler to log passport auth failure

hncj811-va:
ID.me has completed several certificate rotations without proactive notice to Lighthouse.  Unfortunately the SAML Proxy only loads the active certificate at application start up so once the rotation is performed, the responses from ID.me backed IDP's (ID.me, MHV, and DSLogon) fail.

This change logs when an error occurs due to IDP signing key updates.

to force this change line #31 in app.js to:
     (sKeyCert) => !sKeyCert.active 
Then serve the app & attempt id.me login

Resulting example:

{
  "level": "error",
  "message": "Signature check errors: invalid signature: the signature value Bgm5etw36y/SHhVQFZRnnuAprOoYpUXB4g2xZ+vl1EIHA510r1p+VZqB3OszwE9hAlIspSHclukB8C3zQdh7ohYgYPF62R8udMzXocmpcz495j73wbu3vtRDDDoezZZ+HqwQcZIfMDHPTyjSmy/JhyltGXG0lTuchCL26s37RdTdiwlsaZDvZLk5HH1j+EVVh6FQqM9P9d8R295zPshchPT8ei6Bnl4il+BpjNjJVh6DtRulHGP9Ce4FIirD+kWw0OwPILyAZBP1ux7ZP+BEmhK+xcZ/ViuH+IRzrwx1HEod9cd2eIpNvToAZLBsNiL7NXzgpx6hNlMLywKg8oXa8g== is incorrect",
  "request_id": "1e6f7c60-4e26-11ed-801f-09691c6646af",
  "service": "saml-proxy",
  "time": "2022-10-17T14:16:24.647Z"
}
{
  "level": "error",
  "message": 400,
  "request_id": "1e6f7c60-4e26-11ed-801f-09691c6646af",
  "service": "saml-proxy",
  "time": "2022-10-17T14:16:24.648Z"
}

https://vajira.max.gov/browse/API-20473

### DIFF
--- a/src/app.js
+++ b/src/app.js
@@ -27,16 +27,14 @@ const handleMetadata = (argv) => {
       if (metadata.signingKeys) {
         let signingKeyCert;
         signingKeyCert = metadata.signingKeys.find(
-          (sKeyCert) => !sKeyCert.active
+          (sKeyCert) => sKeyCert.active === true
         );
         if (signingKeyCert) {
           //id.me
           argv.spIdpCert = cli.certToPEM(signingKeyCert.cert);
-        } else if (metadata.signingKeys[1]) {
+        } else if (metadata.signingKeys[0]) {
           // normally login.gov
-          argv.spIdpCert = cli.certToPEM(metadata.signingKeys[1].cert);
-          logger.info("Cert misalignment error");
-          logger.info(`Metadata protocol ${metadata.protocol}`);
+          argv.spIdpCert = cli.certToPEM(metadata.signingKeys[0].cert);
         }
       }
       switch (metadata.protocol) {
@@ -121,10 +119,6 @@ function runServer(argv) {
 }
 
 function main() {
-  // let argv = cli.processArgs();
-  // setInterval(() => {
-  //   runServer(cli.processArgs());
-  // }, 3000);
   runServer(cli.processArgs());
 }
 
--- a/src/app.js
+++ b/src/app.js
@@ -27,12 +27,16 @@ const handleMetadata = (argv) => {
       if (metadata.signingKeys) {
         let signingKeyCert;
         signingKeyCert = metadata.signingKeys.find(
-          (sKeyCert) => sKeyCert.active === true
+          (sKeyCert) => !sKeyCert.active
         );
         if (signingKeyCert) {
+          //id.me
           argv.spIdpCert = cli.certToPEM(signingKeyCert.cert);
-        } else if (metadata.signingKeys[0]) {
-          argv.spIdpCert = cli.certToPEM(metadata.signingKeys[0].cert);
+        } else if (metadata.signingKeys[1]) {
+          // normally login.gov
+          argv.spIdpCert = cli.certToPEM(metadata.signingKeys[1].cert);
+          logger.info("Cert misalignment error");
+          logger.info(`Metadata protocol ${metadata.protocol}`);
         }
       }
       switch (metadata.protocol) {
@@ -117,6 +121,10 @@ function runServer(argv) {
 }
 
 function main() {
+  // let argv = cli.processArgs();
+  // setInterval(() => {
+  //   runServer(cli.processArgs());
+  // }, 3000);
   runServer(cli.processArgs());
 }
 
--- a/src/app.js
+++ b/src/app.js
@@ -25,15 +25,14 @@ const handleMetadata = (argv) => {
     if (metadata.protocol) {
       argv.spProtocol = metadata.protocol;
       if (metadata.signingKeys) {
+        // different provider metadata notation requires look up by an active key or falling back to the first entry by default
         let signingKeyCert;
         signingKeyCert = metadata.signingKeys.find(
           (sKeyCert) => sKeyCert.active === true
         );
         if (signingKeyCert) {
-          //id.me
           argv.spIdpCert = cli.certToPEM(signingKeyCert.cert);
         } else if (metadata.signingKeys[0]) {
-          // normally login.gov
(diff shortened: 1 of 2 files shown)


beaugrantham-va:
Mark this PR as ready when it's ready.

src/app.js
```diff
@@ -27,12 +27,16 @@ const handleMetadata = (argv) => {
       if (metadata.signingKeys) {
         let signingKeyCert;
         signingKeyCert = metadata.signingKeys.find(
-          (sKeyCert) => sKeyCert.active === true
+          (sKeyCert) => !sKeyCert.active
```
crolarlibertyva, line 30: You'll need to undo the error inducement.

crolarlibertyva (dismissed):

src/routes/acsHandlers.ts
```diff
@@ -60,6 +60,45 @@ const sufficientLevelOfAssurance = (claims: any) => {
...
   return (req: IConfiguredRequest, res: Response, next: NextFunction) => {
+    const authenticateCB = (arg1: any, arg2: any, err: any, status: any) => {
+      if (err.message) {
+        logger.error(err.message);
+        logger.error(status);
+        return res.redirect("/samlproxy/sp/error");
+      } else {
+        // logger.info(arg2); //user obj like below can be deleted
+        /*
+  "level": "info",
```
dbrown-va, line 71: We can pull out this commented info which is identity provider specific (test data only).

Instead a more generic comment along the lines of 

> 'arg2' contains the user object returned from the SAML identity provider, in this case the happy path should simply pass forward the user identity in the request which be validated later in the execution flow.

src/app.js
```diff
@@ -30,8 +30,10 @@ const handleMetadata = (argv) => {
           (sKeyCert) => sKeyCert.active === true
         );
         if (signingKeyCert) {
+          //id.me
```
dbrown-va, line 33: Recommend pulling out these IDP metadata specific comments out.

Instead a general comment above line 28 can note that different provider metadata notation requires look up by an active or falling back to the first entry by default.

src/routes/acsHandlers.ts
```diff
@@ -60,6 +60,45 @@ const sufficientLevelOfAssurance = (claims: any) => {
 
 export const buildPassportLoginHandler = (acsURL: string) => {
   return (req: IConfiguredRequest, res: Response, next: NextFunction) => {
+    const authenticateCB = (arg1: any, arg2: any, err: any, status: any) => {
+      if (err.message) {
+        logger.error(err.message);
+        logger.error(status);
+        return res.redirect("/samlproxy/sp/error");
+      } else {
```
beaugrantham-va, line 68: `else` is unnecessary here.

beaugrantham-va:
The PR description should describe the change.

It's also helpful for it to link to a Jira user story for additional information.

src/routes/acsHandlers.ts
```diff
@@ -60,6 +60,45 @@ const sufficientLevelOfAssurance = (claims: any) => {
 
 export const buildPassportLoginHandler = (acsURL: string) => {
   return (req: IConfiguredRequest, res: Response, next: NextFunction) => {
+    const authenticateCB = (arg1: any, arg2: any, err: any, status: any) => {
+      if (err.message) {
+        logger.error(err.message);
+        logger.error(status);
+        return res.redirect("/samlproxy/sp/error");
```
crolarlibertyva, line 67: I think this value is a constant is defined somewhere in the req. Should use that.

src/routes/acsHandlers.ts
crolarlibertyva: Would also recommend arg2 be changed to something like 'userInfo'.

src/app.js
hncj811-va: changed

src/routes/acsHandlers.ts
```diff
@@ -60,6 +60,45 @@ const sufficientLevelOfAssurance = (claims: any) => {
 
 export const buildPassportLoginHandler = (acsURL: string) => {
   return (req: IConfiguredRequest, res: Response, next: NextFunction) => {
+    const authenticateCB = (arg1: any, arg2: any, err: any, status: any) => {
```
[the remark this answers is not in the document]
hncj811-va: done

src/routes/acsHandlers.ts
```diff
@@ -60,6 +60,23 @@ const sufficientLevelOfAssurance = (claims: any) => {
 
 export const buildPassportLoginHandler = (acsURL: string) => {
   return (req: IConfiguredRequest, res: Response, next: NextFunction) => {
+    const authenticateCallBack = (
+      arg1: any,
```
dbrown-va, line 64: What is the value of `arg1` in this case?

Recommend a more descriptive name similar to the `userInfo` rename from arg2.

hncj811-va: So the passport authenticate throws a null value for the first argument.  Below is a sample in passport/authenticate.js (line 110) where the code returns the callback following an auth failure.

 return callback(null, false, failures[0].challenge, failures[0].status);

dbrown-va: Got it.  In that case maybe just a rename to `unused`?

dbrown-va: Or it looks like convention is something like `_1: any` since Typescript type enforcement is enabled.  The leading `_` seems to be a decently established convention.

Just will want to run the happy & sad path tests just to be sure.

hncj811-va: Ok just pushed passed all test cases in the regression

crolarlibertyva (approved):

PR #404, opened 2022-11-02
API-15198 deny specific redirect

crolarlibertyva:
Addresses https://vajira.max.gov/browse/API-15198
 - Relay state must begin with '/oauth2'.


src/routes/handlers.js
```diff
@@ -44,11 +44,17 @@ export const samlLogin = function (template) {
         ? authnRequest.relayState
         : req.query?.RelayState;
     }
-    if (relayState == null || relayState == "") {
+    if (
+      relayState == null ||
+      relayState == "" ||
+      !relayState.startsWith("%2Foauth2%2F")
```
dbrown-va, line 50: Confirmed that the invalid redirects via base authorization server navigation are below and will correctly be disallowed.

`RelayState:%2Flogin%2Flogin.htm`

dbrown-va (approved):
Confirmed successful login locally in addition to the relayState path validation.

PR #407, opened 2022-11-22
Toggle styling logging in saml-proxy

hncj811-va:
Logs in the saml-proxy that involve fetching styling elements, such as .css sheets, are distracting to local development and do not add value.
 
This change allows devs to be able to toggle these on and off using a config option. The default leaves them enabled.

https://vajira.max.gov/browse/API-6846

### DIFF
--- a/README.md
+++ b/README.md
@@ -56,6 +56,7 @@ dev-config.json
 ```
 
 `idpAcsUrl`, `idpIssuer`, `idpAudience`, and `idBaseUrl` are all configuration provided by the IDP.
+`styleLogs` is a optional config that can be added to disable all logs related to styling elements like (png, css, woff etc) for local development  (to disable logs set to false)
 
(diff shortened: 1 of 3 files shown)


README.md
```diff
@@ -56,6 +56,7 @@ dev-config.json
 ```
 
 `idpAcsUrl`, `idpIssuer`, `idpAudience`, and `idBaseUrl` are all configuration provided by the IDP.
+`styleLogs` is a optional config that can be added to disable all logs related to styling elements like (png, css, woff etc) for local development  (to disable logs set to false)
```
dbrown-va, line 59: While not a huge issue, `styleLogs` could be misleading that it styles our logging in some way.

I would recommend renaming to `logStyleElementsEnabled` to follow the convention of trailing `Enabled` in most of our new boolean config additions.

dbrown-va (changes requested):
Small rename request to be more descriptive in what the configuration value is doing.

dbrown-va (approved):

PR #412, opened 2022-12-28
API-22644 force upgrade jsonwebtoken

johnmweisz:


package.json
```diff
@@ -143,6 +143,7 @@
     "json-schema": ">=0.4.0",
     "node-forge": ">=1.0.0",
     "ejs": ">=3.1.8",
-    "qs": ">=6.11.0"
+    "qs": ">=6.11.0",
+    "jsonwebtoken": "9.0.0"
```
dbrown-va, line 147: The `>=` syntax used on most of the other resolutions allow for the underlying libraries to upgrade rather that tying to a specific version.  While `9.0.0` is the current highest version, recommend adding the `>=`

dbrown-va (changes requested):
Small bit of feedback, otherwise confirmed the `npm audit` fix.

dbrown-va (approved):

PR #414, opened 2023-01-25
API-21018 hbs alternative ejs

crolarlibertyva:
Addresses https://vajira.max.gov/browse/API-21018
 - Uses ejs for templates instead of hbs. 
 - Removes hbs as a dependency


src/IDPConfig.ts
```diff
@@ -126,10 +126,11 @@ export default class IDPConfig implements IdPOptions {
     res: Response,
     next: NextFunction
   ) {
-    res.render("samlresponse", {
-      AcsUrl: opts.postUrl,
-      SAMLResponse: response.toString("base64"),
-      RelayState: opts.RelayState,
-    });
+    const payload = {
```
dbrown-va, line 129: Any reason this had to be instantiated vs passed directly to the `render` method?

crolarlibertyva: Ables inspection of the payload prior to render call during debug.

dbrown-va (approved):
Confirmed happy path logins as well as failure scenario working when trait lookup fails.

PR #418, opened 2023-02-16
API-000 test users

crolarlibertyva:
Sanity test saml-proxy with multiple happy test users.
 - Currently run manually for now

See https://github.com/department-of-veterans-affairs/lighthouse-saml-proxy/tree/API-000-test-users/test/user_tests#readme for Quick Start

```
cd test/user_tests
npm i
npm test
```


package.json
```diff
@@ -26,8 +26,8 @@
     "start": "node build/app.js",
     "lint": "npx eslint .",
     "lint-fix": "npx eslint . --fix",
-    "test": "jest --testPathIgnorePatterns saml-proxy-regression.test.js",
-    "test:ci": "jest --ci --testPathIgnorePatterns saml-proxy-regression.test.js",
+    "test": "jest --testPathIgnorePatterns test/**/*",
```
dbrown-va, line 29: Any chance we can be more specific on the directories where these are ignored?

I don't want to accidentally disable tests in the future if we split into directories

dbrown-va (approved):
This is a good capability to have since we are moving from vets-api-mockdata to the mock-mpi hosted test users.

PR #420, opened 2023-03-28
API-000 null first last name mpuser

crolarlibertyva:
Preserves original IDP values for null occurrence of either first or last name.

### DIFF
--- a/src/routes/acsHandlers.ts
+++ b/src/routes/acsHandlers.ts
@@ -141,12 +141,12 @@ export const loadICN = async (
     if (first_name) {
       req.user.claims.firstName = first_name;
     } else {
-      logger.warn("Null mpi_user first_name for " + icn);
+      logger.warn("Null mpi_user first_name");
     }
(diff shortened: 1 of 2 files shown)


src/routes/acsHandlers.ts
```diff
@@ -138,8 +138,16 @@ export const loadICN = async (
       result: "success",
     });
     req.user.claims.icn = icn;
-    req.user.claims.firstName = first_name;
-    req.user.claims.lastName = last_name;
+    if (first_name) {
+      req.user.claims.firstName = first_name;
+    } else {
+      logger.warn("Null mpi_user first_name for " + icn);
```
dbrown-va, line 144: We should be able to correlate the logins without needing the ICN logged to minimize direct PII.

Most full Auth Code logins are quite spread out throughout the day.

dbrown-va (approved):

dbrown-va:
Confirmed log level

PR #421, opened 2023-03-30
Api 13580 eslint plugin jsdoc

dmekala-va:


dmekala-va:
Link to JIRA story: https://vajira.max.gov/browse/API-13580

dmekala-va:
This screenshot shows it is running eslint jsdoc with no warnings

Screenshot 2023-03-30 at 2 57 16 PM

dbrown-va (changes requested):
Since the SAML Proxy is combined repo of .js and .ts files, we may have to adjust to ensure the `.ts` methods are also covered.  It seems like the `.ts` extension files aren't getting flagged.

dmekala-va:
Picture below shows the functionality that the .ts files are getting flagged now after adding some more plugin logic to the .estlintrc.yml file. These warnings have been fixed so will not appear anymore after running script. 

Screenshot 2023-04-04 at 9 02 48 AM

dbrown-va (approved):
Thanks for addressing the feedback on the typescript files as well as the extra jsdocs to allow for the warnings when fully missing.  Looks good.

PR #423, opened 2023-04-14
API-13875-fix-modify-regression-test

mwise-va:
- Added TIMEOUT env variable, so we can easily adjust regression test timeout within param store. 
- Added TIMEOUT to the "modify" regression test, so we can increase as needed.


dbrown-va (approved):

dbrown-va:
Confirmed `/dvp/common/ecs-saml-proxy/codebuild/regression_test_timeout` parameter set.

PR #426, opened 2023-04-17
Api 25720 Typescript Errors

dmekala-va:
This PR is to resolve the typescript errors that exist in various files.

Link to JIRA: https://vajira.max.gov/browse/API-25720

### DIFF
--- a/test/mockMpiUserClient.ts
+++ b/test/mockMpiUserClient.ts
@@ -6,12 +6,12 @@ export default class MockMpiUserClient {
 
   public async getMpiTraitsForLoa3User(
     user: SAMLUser
-  ): Promise<{ icn: string; first_name: string; last_name: string }> {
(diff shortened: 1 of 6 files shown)


test/mockMpiUserClient.ts
```diff
@@ -1,23 +1,24 @@
...
 
   public async getMpiTraitsForLoa3User(
     user: SAMLUser
   ): Promise<{ icn: string; first_name: string; last_name: string }> {
     if (this.findUserInMVI) {
       return {
         icn: "123",
-        first_name: user.firstName,
-        last_name: user.lastName,
+        first_name: user.firstName!,
```
johnmweisz, line 13: instead of the non-null assertion operator you can change the expected type to be 
`Promise<{ icn: string; first_name: string | undefined; last_name: string | undefined }`

johnmweisz (approved):

dbrown-va (approved):

PR #430, opened 2023-05-01
API-25935 Request Vulnerability

dmekala-va:
This PR is for removing the request and request promise native vulnerabilities that exist. Since the library was deprecated, this PR replaces the deprecated library with Axios http lib.

Link to JIRA: https://vajira.max.gov/browse/API-25935


src/MpiUserClient.ts
```diff
@@ -1,5 +1,6 @@
-import * as request from "request-promise-native";
 import { SAMLUser } from "./SAMLUser";
+import axios from "axios";
+//import * as request from "request-promise-native";
```
crolarlibertyva, line 3: Cleanup needed.

src/idpMetadata.js
```diff
@@ -58,92 +57,102 @@ export function fetch(url) {
...
+                logger.error("unable to parse RoleDescriptor metadata", e);
+              }
+            }
+            return resolve(metadata);
+          });
         }
-        return resolve(metadata);
+      })
+      .catch(() => {
+        console.log("error occurred");
```
crolarlibertyva, line 155: Use a logger here. Eg. ```logger.error(". . . ."```

crolarlibertyva (changes requested):
All is good for me. Just a couple of little things to correct.

crolarlibertyva (approved):

PR #434, opened 2023-05-22
API-25662 adds InResponseTo SAML attribute

johnmweisz:
Includes InResponseTo property in idp SAML response to support Okta update.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -134,6 +134,7 @@ sp-metadata.xml
 
 # Local config
 dev-config.json
+local-config.json
 
 ## Webstorm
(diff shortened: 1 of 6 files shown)


mwise-va (dismissed):
Reviewed changes and tested locally. I was able to login and see that "inResponseTo"  was added.

dbrown-va:
This would be good to have a unit test validating it pulling the ID from the session for the `inResponseTo` field.

crolarlibertyva (dismissed):

crolarlibertyva (approved):

PR #435, opened 2023-05-26
API-000 InResponseTo Fix

johnmweisz:
- This pr will allow us to deploy InResponseTo assertion to all environments without breaking existing flow.
- The end result will be that sometimes Sandbox and Production (environments with more than one task) will send InResponseTo to Okta and sometimes it wont.  It will no longer send an incorrect fallback value.
- A session fix will be required that moves session to a shared store such as Redis.


crolarlibertyva (commented):
Replay regression test is failing for me.

crolarlibertyva (approved):

PR #439, opened 2023-06-07
API-27106 forward SAMLRequest ID

johnmweisz:
- Reverts session related changes
- Forwards Okta ID to idp and returns the InResponseTo value to Okta.
- Unit tests check that InResponseTo is added to SAMLResponse if available.
- This is a necessary change to be compliant with SAML 2.0 standards and support the Okta upgrade.


johnmweisz:
wanted to verify that saml decode is stable, I think existing flow does not require inflate but will keep this in mind if we see any errors as it is properly logged. https://stackoverflow.com/questions/24829612/working-with-a-compressed-base64-query-parameter-with-nodejs

src/routes/acsHandlers.ts
```diff
@@ -306,6 +305,7 @@ export const serializeAssertions = (
     logRelayState(req, logger, "to Okta");
   }
   authOptions.authnContextClassRef = req.user.authnContext.authnMethod;
+  if (inResponseTo) authOptions.inResponseTo = inResponseTo;
```
dbrown-va, line 308: For a future story, consider erroring when `inResponseTo` is missing rather than deferring to Okta error handling (which relays back to a SAML proxy 500)

dbrown-va (approved):

PR #452, opened 2023-07-17
Log injection Fix

mwise-va:
**Description:**
saml-proxy logging of relay state may warrant some level of relay state validation if not already present.

**Related Issues:**
https://jira.devops.va.gov/browse/API-28193

**A/C:**
CodeQL log injection alert's not showing up on lighthouse-saml-proxy repo.

### DIFF
--- a/src/utils.js
+++ b/src/utils.js
@@ -9,6 +9,7 @@ import fs from "fs";
 export function getPath(path) {
   return path.startsWith("/") ? path : "/" + path;
 }
+
 /**
  * This function creates a check to see if host is equal to localhost then
  * returns a req url accordingly
@@ -28,6 +29,7 @@ export function getReqUrl(req, path) {
     )}`;
   }
 }
+
 /**
  * Creates a check to remove headers with BEGIN and END message
  *
@@ -41,6 +43,7 @@ export function removeHeaders(cert) {
   }
   return cert;
 }
+
 /**
  * This function logs the relay state
  *
--- a/src/utils.js
+++ b/src/utils.js
@@ -39,7 +39,7 @@ export function getReqUrl(req, path) {
 export function removeHeaders(cert) {
   const pem = /-----BEGIN (\w*)-----([^-]*)-----END (\w*)-----/g.exec(cert);
   if (pem && pem.length > 0) {
-    return pem[2].replace(/[\n|\r\n]/g, "");
+    return sanitize(pem[2]);
   }
   return cert;
 }
--- a/src/utils.js
+++ b/src/utils.js
@@ -49,17 +49,13 @@ export function removeHeaders(cert) {
  * @param {*} step relay state step
  */
 export function logRelayState(req, logger, step) {
+  let relayStateBody = req.body.RelayState;
+  let relayStateQuery = req.query.RelayState;
   if (req.body.RelayState) {
-    let relayStateBody = req.body.RelayState.replace(/\n|\r/g, "");;
-  }
-  else {
-    let relayStateBody = req.body.RelayState;
+    relayStateBody = req.body.RelayState.replace(/\n|\r/g, "");
   }
   if (req.query.RelayState) {
-    let relayStateQuery = req.query.RelayState.replace(/\n|\r/g, "");
-  }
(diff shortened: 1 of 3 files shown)


src/utils.js
```diff
@@ -42,25 +42,31 @@ export function removeHeaders(cert) {
...
+  });
+}
+
+/**
+ * This function sanatizes a message, by replacing new line escapes with ""
+ *
+ * @param {*} message message that needs sanitized
+ * @returns {*} returns sanitized message
+ */
+export function sanatize(message) {
```
beaugrantham-va, line 68: s/sanatize/sanitize

src/utils.js
```diff
@@ -42,25 +42,31 @@ export function removeHeaders(cert) {
...
+    step: step,
+    session: req.sessionID,
+  });
+}
+
+/**
+ * This function sanatizes a message, by replacing new line escapes with ""
+ *
+ * @param {*} message message that needs sanitized
+ * @returns {*} returns sanitized message
```
beaugrantham-va, line 66: The param and return types are known here (string) and should be used instead of *.

src/utils.js
beaugrantham-va, line 68: This will throw an unhandled exception if the user provided value is null or undefined.

As simple as this is, it should have an accompanying unit test.

src/utils.js
```diff
@@ -42,25 +42,31 @@ export function removeHeaders(cert) {
...
-      relayStateBody,
-      relayStateQuery,
-      step: step,
-      session: req.sessionID,
-    }
-  );
+  const logMessage = `Relay state ${step} - body: ${req.body.RelayState} query: ${req.query.RelayState}`;
+  logger.info(sanatize(logMessage), {
+    time: new Date().toISOString(),
+    relayStateBody: req.body.RelayState,
```
beaugrantham-va, line 55: Why don't these additional properties also need to be sanitized?

src/utils.js
mwise-va: Typo fixed and function is now named "sanitize".

src/utils.js
mwise-va: Corrected return type. 

 * @returns {string} returns sanitized message

src/utils.js
mwise-va: - Added exception handling for null and undefined input's.
- Added (2) unit tests.
1. Tests for new line/carriage returns input's to be sanitized correctly.
2. Tests for exception handling.

src/utils.js
mwise-va: Added sanitize() to logRelayState properties.

src/utils.js
```diff
@@ -41,26 +43,39 @@ export function removeHeaders(cert) {
...
+}
+
+/**
+ * This function sanitizes a message, by replacing new line escapes with ""
+ *
+ * @param {*} message message that needs sanitized
+ * @returns {string} returns sanitized message
+ */
+export function sanitize(message) {
+  if (message === null || message === undefined) {
```
crolarlibertyva, line 75: If (message) (truthy) would work here.

mwise-va: Added truthy sanitize check.

src/utils.js
```diff
@@ -37,30 +39,44 @@ export function getReqUrl(req, path) {
...
+    relayStateQuery: relayStateQuery,
+    step: relayStateStep,
+    session: req.sessionID,
+  });
+}
+
+/**
+ * This function sanitizes a message, by replacing new line escapes with ""
+ *
+ * @param {*} message message that needs sanitized
```
crolarlibertyva, line 71: This is assumed to be a string right?

src/utils.js
crolarlibertyva: @parms message needs correction still.

mwise-va: Resolved.

 * @param {string} message message that needs sanitized

src/utils.js
mwise-va: Resolved. 
 
 * @param {string} message message that needs sanitized

crolarlibertyva (approved):

PR #454, opened 2023-07-18
API-28194 codeql missing csrf

dmekala-va:
This PR addresses this codeql alert scan:
https://github.com/department-of-veterans-affairs/lighthouse-saml-proxy/security/code-scanning/8

Link to Jira: https://jira.devops.va.gov/browse/API-28194


src/routes/index.js
```diff
@@ -157,6 +157,15 @@ export default function configureExpress(
     })
   );
   app.use(flash());
+  app.use(lusca.xframe("SAMEORIGIN"));
+  app.use(lusca.p3p("ABCDEF"));
+  app.use(lusca.hsts({ maxAge: 31536000 }));
+  app.use(lusca.xssProtection(true));
+  app.use(
```
johnmweisz, line 164: I think there should only be one app.use
consider doing as described below from the documentation:
app.use(lusca({
    csrf: true,
    csp: { /* ... */},
    xframe: 'SAMEORIGIN',
    p3p: 'ABCDEF',
    hsts: {maxAge: 31536000, includeSubDomains: true, preload: true},
    xssProtection: true,
    nosniff: true,
    referrerPolicy: 'same-origin'
}));

src/routes/index.js
```diff
@@ -21,7 +21,7 @@ import { getParticipant } from "./handlers";
 
 import promBundle from "express-prom-bundle";
 import * as Sentry from "@sentry/node";
-
+import lusca from "lusca";
```
dbrown-va, line 24: Confirmed we are using `express-session` which is called out as a requirement in the lusca docs.

src/routes/index.js
```diff
@@ -157,6 +157,15 @@ export default function configureExpress(
     })
   );
   app.use(flash());
+  app.use(lusca.xframe("SAMEORIGIN"));
+  app.use(lusca.p3p("ABCDEF"));
+  app.use(lusca.hsts({ maxAge: 31536000 }));
+  app.use(lusca.xssProtection(true));
+  app.use(
+    lusca({
+      csrf: false,
```
dbrown-va, line 166: I would have expected `csrf` protection to have been enabled here.  But perhaps I am misunderstanding the lusca docs.

src/routes/index.js
```diff
@@ -157,6 +157,15 @@ export default function configureExpress(
     })
   );
   app.use(flash());
+  app.use(lusca.xframe("SAMEORIGIN"));
```
dbrown-va, line 160: The various lusca docs should be consolidated to a single entry w/ config, similar to the session configuration above.

```
app.use(lusca({
    csrf: true,
    xframe: 'SAMEORIGIN',
    p3p: 'ABCDEF',
    hsts: {maxAge: 31536000},
    xssProtection: true,
}));
```

dbrown-va (changes requested):
Would like to see the `app.use` statements consolidated to a single statement with config as noted in the reviewer.

src/routes/index.js
dmekala-va: after trial and error I found that csrf being set to false is the only fix that works. I did try setting to true but it was failing the csrf fix and saml-proxy happy path. It wasn't finding the csrf token

johnmweisz: If this does rely on session, I would expect it to be unstable as we are using the in memory local development version of session which is not shared across tasks.  We also do a lot of redirecting to other sites, which would also impact the stability.

johnmweisz: Again, I'm concerned about this being necessary and stable.

crolarlibertyva: Bullet bitten. Confirmed the sessions are winding up in the redis cache.

johnmweisz: removing session is probably better than using redis

dmekala-va: reverted redis change. was failing build

crolarlibertyva:
Heads up. Working the session aspect in a different branch.https://github.com/department-of-veterans-affairs/lighthouse-saml-proxy/tree/API-28194-redis-session.

PR #458, opened 2023-08-03
API-28822 remove session

johnmweisz:
- Removes session dependency
- Abstracts ID and ReplayState retrieval
- Removes some repetitive req var stashing.

### DIFF
--- a/src/routes/acsHandlers.ts
+++ b/src/routes/acsHandlers.ts
@@ -5,6 +5,7 @@ import {
   accessiblePhoneNumber,
   sanitize,
   getSamlId,
+  getRelayState,
 } from "../utils";
 import { ICache, IConfiguredRequest } from "./types";
 import { preparePassport } from "./passport";
@@ -88,7 +89,7 @@ export const buildPassportLoginHandler = (acsURL: string) => {
       (req.body && req.body.SAMLResponse)
     ) {
       const ssoResponse = {
-        state: req.query.RelayState || req.body.RelayState,
+        state: getRelayState(req),
         url: getReqUrl(req, acsURL),
       };
       if (req.options) {
@@ -224,7 +225,7 @@ export const testLevelOfAssuranceOrRedirect = (
     req.user.claims &&
     !sufficientLevelOfAssurance(req.user.claims)
   ) {
-    if (!req.query?.RelayState && !req.body?.RelayState) {
+    if (!getRelayState(req)) {
       throw {
         message: "Error: Empty relay state during loa test. Invalid request.",
         status: 400,
@@ -235,7 +236,7 @@ export const testLevelOfAssuranceOrRedirect = (
         pathname: SP_VERIFY,
         query: {
           authnContext: "http://idmanagement.gov/ns/assurance/loa/3",
-          RelayState: req.query?.RelayState || req.body?.RelayState,
+          RelayState: getRelayState(req),
         },
       })
     );
--- a/src/routes/acsHandlers.ts
+++ b/src/routes/acsHandlers.ts
@@ -5,6 +5,7 @@ import {
   accessiblePhoneNumber,
   getSamlId,
   getRelayState,
+  sanitize,
 } from "../utils";
 import { ICache, IConfiguredRequest } from "./types";
 import { preparePassport } from "./passport";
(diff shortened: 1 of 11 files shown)


src/routes/acsHandlers.ts
```diff
@@ -247,37 +243,33 @@ export const testLevelOfAssuranceOrRedirect = (
 export const validateIdpResponse = (cache: ICache, cacheEnabled: Boolean) => {
   return async (req: IConfiguredRequest, res: Response, next: NextFunction) => {
     if (cacheEnabled) {
-      const sessionIndex = req?.user?.authnContext?.sessionIndex;
```
johnmweisz, line 250: We could continue to use this but I would prefer to use the same identifier that Okta uses, potential for discussion.

src/routes/acsHandlers.ts
```diff
@@ -247,37 +243,33 @@ export const testLevelOfAssuranceOrRedirect = (
...
-      if (!sessionIndex) {
-        logger.error("No session index found in the saml response.");
+      const samlId = getSamlId(req);
+      if (!samlId) {
```
johnmweisz, line 247: This is a notable change, we are currently letting requests without InResponseTo fail at Okta, this shifts that left and logs.  We are currently not receiving any requests without this value, so practically not a significant change, but I would rather be explicit in the codebase that this value is important instead of having it be indirectly enforced by Okta.

src/utils.js
```diff
@@ -177,3 +180,79 @@ export function sassMiddleware(options) {
...
+ * Gets the ID (InResponseTo) from SAMLRequest or SAMLResponse
+ *
+ * @param {IConfiguredRequest} req The HTTP request
+ * @returns {*} a string if ID or InResponseTo is present
+ */
+export function getSamlId(req) {
+  if (req?.authnRequest?.id) {
+    return req?.authnRequest?.id;
+  }
+  if (req?.body?.SAMLResponse) {
```
mwise-va, line 194: if/else if/else could clean up this function, instead of having a bunch of if checks.

crolarlibertyva (commented):
Confirmed that all regression tests pass, except for 'modify' as expeted.

src/utils.js
```diff
@@ -177,3 +180,79 @@ export function sassMiddleware(options) {
...
+ * @param {string} samlRequest the raw samlRequest
+ * @returns {*} a string if ID is present
+ */
+function getIdToFromSAML(samlRequest) {
+  try {
+    const decoded = Buffer.from(samlRequest, "base64").toString("ascii");
+    const parser = new DOMParser();
+    return parser
+      .parseFromString(decoded)
+      ?.documentElement?.getAttributeNode("ID")?.nodeValue;
```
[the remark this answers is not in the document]
johnmweisz: undefined (falsy)

src/utils.js
johnmweisz: Hmm not sure how I would do that without changing the logic.. I want it to fall through if false/falsy

src/utils.js
```diff
@@ -177,3 +180,79 @@ export function sassMiddleware(options) {
...
+ * Retrieves RelayState from Request
+ *
+ * @param {IConfiguredRequest} req the raw request
+ * @returns {*} a string if state is present
+ */
+export function getRelayState(req) {
+  if (req?.authnRequest?.RelayState) {
+    return req?.authnRequest?.RelayState;
+  }
+  if (req?.query?.RelayState) {
```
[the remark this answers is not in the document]
johnmweisz: Same - not sure how I would do that without changing the logic.. I want it to fall through if false/falsy

src/utils.js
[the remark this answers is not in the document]
johnmweisz: I do not want a placeholder value to be used to make it clear the value was not found but also don't want to throw an exception though I could and just add try/catch to ignore it in the code

src/routes/acsHandlers.ts
dbrown-va: This is good, approve that shift left rather than relying on Okta.

dbrown-va (approved):

src/utils.js
dbrown-va: This could be if/else if/else if (no else).  But functionally would be the same result since each of those if's include a return.

PR #467, opened 2023-08-31
API-29232 okta as an idp with test users

mwise-va:
**Summary**

Implementation of saml-proxy acting as a Service Provider (SP) and Okta acting as an Identity Provider (IDP) with test user accounts.

**Related Issue(s)**

https://vajira.max.gov/browse/API-29232

**Testing done**

- Successfully logged into Okta using the new Mock IDP button and a test user account, while saml-proxy acts as a SP and Okta is the IDP that stores the test users.
- After logging in I was able to see user info at - http://localhost:8080/userdetails
- All regression test passed.
- All unit test passed.

**Screenshots**

Screenshot 2023-08-29 at 7 45 14 PM

Screenshot 2023-08-30 at 10 07 25 PM

Screenshot 2023-08-30 at 10 08 14 PM

Screenshot 2023-08-30 at 10 07 43 PM

Screenshot 2023-08-30 at 9 50 32 PM


src/routes/passport.js
```diff
@@ -76,6 +76,15 @@ export function preparePassport(strategy) {
 export const selectPassportStrategyKey = (req) => {
   const origin = req.headers.origin;
   let passportKey = "id_me";
+  if (origin === "null") {
```
johnmweisz, line 79: Any idea why origin us null only for Okta? Custom code for okta makes me a little nervous.

views/login_selection.ejs
```diff
@@ -50,6 +50,9 @@
       </svg>
     </a>
     <% } %>
+    <% if (mock_idp_enabled) { %>
+      <a class="no-external-icon usa-button mockidp-signin" href="<%- mock_idp_login_link -%>">Mock IDP</a>
```
johnmweisz, line 54: Moving to draft as this is blocked by the spike and I don't think we would want this named "Mock IDP"

crolarlibertyva: I'd like modifications to the template to be more config driven. This was good for proving out a concept. But it would be better to address this as tech debt too.

src/routes/passport.js
johnmweisz: Agreed, if we find the logic doesn't scale it should be replaced, I think it will end up being a decode SAML and get attribute string match type fix as shown here, removing the obscure dependency on the origin header will improve stability as this common header is something that can be added or removed based on CSRF support.

crolarlibertyva:
Might be cool in the future. LOL!!

PR #468, opened 2023-09-05
API-10892 techdebt passport selection fix

crolarlibertyva:
Relates to API-10892
 - Address tech debt risk. login.gov is the only idp populated with headers.origin. 
 - Replaces a function that won't work for any other IDP


mwise-va (dismissed):
Code changes look good and will support adding new IDP's to saml-proxy in the future. It was a good idea to add the decoding of the SAML response and returning the issuer to the utils. 

Unit test and regression test all passed. Also tested a login using the sample app with id.me

src/routes/passport.js
```diff
@@ -74,12 +75,21 @@ export function preparePassport(strategy) {
  * @returns {*} A string with the correct spIdp key
  */
 export const selectPassportStrategyKey = (req) => {
-  const origin = req.headers.origin;
-  let passportKey = "id_me";
-  Object.entries(req.sps.options).forEach((spIdpEntry) => {
-    if (spIdpEntry[1].idpSsoUrl.startsWith(origin)) {
-      passportKey = spIdpEntry[0];
-    }
+  const samlResponse = req.body?.SAMLResponse
```
johnmweisz, line 78: could use short circuit eval here..
`const samlResponse = req.body?.SAMLResponse || req.query?.SAMLResponse`

src/routes/passport.js
```diff
@@ -74,12 +75,21 @@ export function preparePassport(strategy) {
...
+    ? req.body.SAMLResponse
+    : req.query.SAMLResponse;
+  const issuer = issuerFromSamlResponse(samlResponse);
+  const spIdpKeys = Object.keys(req.sps.options);
+  const foundSpIdpKey = spIdpKeys.find((spIdpKey) => {
+    const url = new URL(req.sps.options[spIdpKey].idpMetaUrl);
+    const domain_parts = url.host.split(".");
+    const domain = (domain_parts.length > 1
```
johnmweisz, line 86: As this is a 'URL' we might be able to use a property here (ex url.hostname) instead of this splitting logic https://developer.mozilla.org/en-US/docs/Web/API/URL#Properties

src/routes/passport.js
```diff
@@ -74,12 +75,21 @@ export function preparePassport(strategy) {
...
+      ? domain_parts[domain_parts.length - 2] +
+        "." +
+        domain_parts[domain_parts.length - 1]
+      : domain_parts[0]
+    ).replace("preview", ""); // An IDP broke convention and created a mismatch between the metdata url domain vs the issuer in the saml response
```
johnmweisz, line 91: makes me nervous to see preview specific logic as this is not used in deployed env.

src/utils.js
```diff
@@ -220,6 +220,29 @@ function getInResponseToFromSAML(samlResponse) {
...
+ */
+export function issuerFromSamlResponse(samlResponse) {
+  try {
+    const decoded = Buffer.from(samlResponse, "base64").toString("ascii");
+    const parser = new DOMParser();
+    const issuerElems = parser
+      .parseFromString(decoded)
+      .documentElement.getElementsByTagNameNS(
+        "urn:oasis:names:tc:SAML:2.0:assertion",
+        "Issuer"
```
johnmweisz, line 237: I was wondering if the pretext mattered (saml:... vs saml2:...) I guess not, good to know!

src/routes/passport.js
johnmweisz: maybe not I guess hostname is basically the same as host

johnmweisz: you could use url.origin which should match the first part of the issuer exactly

src/routes/passport.js
johnmweisz: I do wonder how Okta issuer works, if they use '..Okta...' as the issuer for all env regardless of hosted domain, this could be a problem.

src/routes/passport.js
johnmweisz: So it seems that matching origin against issuer might not work directly as IDP's like okta have a different issuer than the domain, that's why you are adding extra logic here... I'm wondering if this is more brittle than we thought.

johnmweisz: Maybe this needs to be a configurable item? each IDP has an "issuer" property or something in the config.

crolarlibertyva: As it stands, only login.gov populates that. No other IDP populates that.

src/routes/passport.js
crolarlibertyva: I presume they do. Would have to experiment with sandbox okta to verify. I doubt they would ever be selected as one of the IDP choices anyways.

src/routes/passport.js
crolarlibertyva: Looked more into trying to utilize origin. Still wind up having to isolate the domain portion in order to be guaranteed a search match. Eg. issuer might be www.xxx.yyy and metadata url might be idp.xxx.yyy.

src/routes/passport.js
```diff
@@ -74,12 +75,31 @@ export function preparePassport(strategy) {
...
+  return foundSpIdpKey ? foundSpIdpKey : spIdpKeys[0];
 };
+
+/**
+ * Returns the domain that cooresponds to the host of the metadata url.
+ *
+ * @param {spConfig} spIdpOption The SP Config Option
+ * @returns {string} The domain name
+ */
+function spIdpHostDomain(spIdpOption) {
```
mwise-va, line 97: Would this better fit in utils.js?

crolarlibertyva: So far no. This is pretty specific to the calling function. This function isn't even currently exported.

src/routes/passport.js
```diff
@@ -74,12 +75,31 @@ export function preparePassport(strategy) {
...
+  const spIdpKeys = Object.keys(req.sps.options);
+  const foundSpIdpKey = spIdpKeys.find((spIdpKey) => {
+    const spIdpOption = req.sps.options[spIdpKey];
+    const domain = spIdpHostDomain(spIdpOption);
+    return spIdpOption.idpIssuerMatchOverride
+      ? issuer.includes(spIdpOption.idpIssuerMatchOverride)
+      : issuer.includes(domain);
   });
-  return passportKey;
+  return foundSpIdpKey ? foundSpIdpKey : spIdpKeys[0];
```
johnmweisz, line 88: Want to make sure I understand this: so this would mean if a spkey isn't found use default spIdpKeys[0] which I'm guessing is supposed to be ID.me? (This logic does appear to be consistent with what exists but might want to hard code "id_me" instead of using spIdpKeys[0])

johnmweisz (commented):
looks good, as this is a core change would like tech lead approval as well.

src/routes/passport.js
crolarlibertyva: The default is whatever is first in the list, ie. outside of the "idpSamlLogins" in the config. It's currently ID.me but if we were to switch where the login.gov config lives vs Id.me, then login.gov would be the default. Not that that would ever happen.

crolarlibertyva: Also, the first entry is guaranteed to be there.  Saml proxy won't deploy without that first IDP entry filled out.

crolarlibertyva:
SAML Proxy will be deprecated and replaced.

PR #469, opened 2023-09-06
API-000 Add Lodash Modules

mwise-va:
**Decsription:** 
The following Lodash modules were added to saml-proxy package.json
1.     "lodash.includes": "^4.3.0",
2.     "lodash.isboolean": "^3.0.3",
3.     "lodash.isinteger": "^4.0.4",
4.     "lodash.isnumber": "^3.0.3",
5.     "lodash.isplainobject": "^4.0.6",
6.     "lodash.once": "^4.1.1"

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -18,8 +18,11 @@
         "@department-of-veterans-affairs/formation": "^8.0.1-3",
         "uswds": "^1.6.14",
         "@sentry/node": "^6.4.0",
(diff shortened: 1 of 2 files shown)


dbrown-va (changes requested):
Small adjustment to dependency order in package.json

crolarlibertyva (approved):
- Regression tested with docker-compose.
- Also tested with 
```docker run --rm testandlint npm run-script test:ci```

PR #483, opened 2024-02-13
API-34013 ICN error regression test fix

dmekala-va:
This PR is to temporarily alter the icn_error regression test to add a separate env password var so that it can use the old account's password instead of relying on the new id.me accounts. The reason for this change is because there is no valid user in the new id.me test accounts for the icn_error test to use.

Link to Jira: https://jira.devops.va.gov/browse/API-34013

### DIFF
--- a/cicd/buildspec-deploy.yml
+++ b/cicd/buildspec-deploy.yml
@@ -49,6 +49,7 @@ env:
     USER_PASSWORD: "/dvp/common/ecs-saml-proxy/codebuild/user_password"
     VALID_USER_EMAIL: "/dvp/common/ecs-saml-proxy/codebuild/valid_user_email"
     ICN_ERROR_USER_EMAIL: "/dvp/common/ecs-saml-proxy/codebuild/icn_error_user_email"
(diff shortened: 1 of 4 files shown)


cicd/buildspec-deploy.yml
```diff
@@ -49,6 +49,7 @@ env:
     USER_PASSWORD: "/dvp/common/ecs-saml-proxy/codebuild/user_password"
     VALID_USER_EMAIL: "/dvp/common/ecs-saml-proxy/codebuild/valid_user_email"
     ICN_ERROR_USER_EMAIL: "/dvp/common/ecs-saml-proxy/codebuild/icn_error_user_email"
+    ICN_ERROR_PASSWORD: "/dvp/common/ecs-saml-proxy/codebuild/icn_error_password"
```
mwise-va, line 52: I checked DSVA AWS Parameter Store for `/dvp/common/ecs-saml-proxy/codebuild/icn_error_password` and couldn't find it.

Looks like we will need to add this variable to AWS Parameter Store before merging.

mwise-va: I'd add it to param store first, since this PR relies on that variable being present.

PR #490, opened 2024-07-02
API-36203 banner data model

dmekala-va:
This PR adds a data model and accessor class for the saml proxy banner. The banner.js represents the banner data model with banner parameters. And the BannerService implements the CRUD operations for banners in a Redis store.

Link to Jira: https://jira.devops.va.gov/browse/API-36203


src/bannerService.ts
```diff
@@ -0,0 +1,105 @@
...
+import Redis, { RedisClient } from "redis";
+
+interface IRedisClient {
+  setAsync(key: string, value: string): Promise<void>;
+  getAsync(key: string): Promise<string | null>;
+  delAsync(key: string): Promise<void>;
+  keysAsync(pattern: string): Promise<string[]>;
+}
+
+class RedisClientAdapter implements IRedisClient {
```
johnmweisz, line 12: this might be able to go away, I would look at the existing cache object referenced in another comment.

src/bannerService.ts
```diff
@@ -0,0 +1,105 @@
...
+  }
+
+  async delAsync(key: string): Promise<void> {
+    const delAsync = promisify<string, number>(this.client.del).bind(
+      this.client
+    );
+    return delAsync(key).then(() => {});
+  }
+
+  async keysAsync(pattern: string): Promise<string[]> {
```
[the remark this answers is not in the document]
dmekala-va: It retrieves all the keys from the redid store. I was thinking this one could be used to get/delete all the banners but could delete this and just create a new getBanners function instead

dmekala-va:
Comments addressed

src/banner.ts
```diff
@@ -0,0 +1,35 @@
...
+  startTime: Date;
+  endTime: Date;
+  message: string;
+  enabled: boolean;
+  priority: number;
+  order: number;
+
+  constructor(
+    id: string,
+    startTime: Date,
```
johnmweisz, line 12: this might be better as a timestamp, though we can fine tune this later.

src/bannerService.ts
```diff
@@ -0,0 +1,84 @@
...
+      console.error("Error retrieving banner:", error);
+      return null;
+    }
+  }
+
+  //can only use the get if we want to retreive individual banners when you know IDS
+  //we will need a new keys function to get all banners without knowing the IDS
+  async getBanners(): Promise<Banner[]> {
+    try {
+      const keys = await this.cache.keys("banner:");
```
johnmweisz, line 41: I thought there is a way to query for partial key value without using keys but it's been a while, can you confirm if this can be done with a doc search or scan or something instead?

src/bannerService.ts
```diff
@@ -0,0 +1,84 @@
...
+      return banners;
+    } catch (error) {
+      console.error("Error retrieving banners from cache:", error);
+      return [];
+    }
+  }
+
+  async createBanner(banner: Banner): Promise<Banner | null> {
+    try {
+      const key = this.generateKey(banner.id);
```
johnmweisz, line 56: Just want to clarify that this means the service layer will need to generate the banner.id value?  You could probably do that here instead since the data layer cares about accidentally creating duplicate key values but the service probably doesn't.  With only a few entries at a time, duplicated values is probably so unlikely its not important but this could be a weird error some day.

src/bannerService.ts
```diff
@@ -0,0 +1,84 @@
...
+      });
+      const banners = await Promise.all(bannerPromises);
+      return banners;
+    } catch (error) {
+      console.error("Error retrieving banners from cache:", error);
+      return [];
+    }
+  }
+
+  async createBanner(banner: Banner): Promise<Banner | null> {
```
johnmweisz, line 54: Do we know the TTL on this?  I think using the existing cache was on paper the fastest solution but we have all kinds of edge cases that could bite us.

src/bannerService.ts
```diff
@@ -0,0 +1,84 @@
...
+      await this.cache.set(key, value);
+    } catch (error) {
+      console.error("Error updating banner:", error);
+    }
+  }
+
+  async deleteBanner(id: string): Promise<void> {
+    try {
+      const key = this.generateKey(id);
+      await this.cache.delete(key);
```
johnmweisz, line 79: Do we know what cache.delete returns?  We might need this to return something to generate an appropriate status code.  For next story probably but worth thinking about now especially if it impacts unit tests.

src/bannerService.ts
```diff
@@ -0,0 +1,84 @@
...
+      console.error("Error creating banner:", error);
+      return null;
+    }
+  }
+
+  async updateBanner(id: string, banner: Banner): Promise<void> {
+    try {
+      const key = this.generateKey(id);
+      const value = JSON.stringify(banner);
+      await this.cache.set(key, value);
```
johnmweisz, line 70: Do we know what cache.set returns? We might need this to return something to generate an appropriate status code. For next story probably but worth thinking about now especially if it impacts unit tests.

src/bannerService.ts
dmekala-va: Yes we could do a redid scan operation like below:
`SCAN cursor [MATCH pattern] [COUNT count]`

That would get rid of the need for the keys but thought it would be neater to add a keys function instead, and keys is useful enough where it could be used elsewhere too. If we prefer this way though I can change it

src/bannerService.ts
dmekala-va: I guess in this case we could refine a bit more and add a check saying generate a banner id if it's not already set?

dmekala-va: Instead of having generateKey() we could have something like generateUniqueId() where a UUID could be generated instead of a generic banner id

src/bannerService.ts
dmekala-va: Just after researching online it looks like the TTL is assumed to be 3600 secs for cached items usually. If we want we could add logic to set the TTL in createBanner()

src/bannerService.ts
dmekala-va: I think the cache.set is the same way it usually returns a boolean value or a promise that resolves to a boolean. Will add this

src/bannerService.ts
dmekala-va: Actually looking to implement it right now, I don't think there is a way to use the scan function without modifying the RedisCache/ICache to implement "scan" first. So if we wanted to do scan, we would still have to add a scan function in types.ts since bannerService.ts is only using this existing cache instance

dmekala-va: Looking at this again and I think scan makes things a bit more complicated, I feel like to be consistent and staightforward the keys function makes better sense to me personally. I think scan the only benefit of scan is that it is more suitable for larger datasets.

src/banner.ts
dmekala-va: implemented

src/bannerService.ts
dmekala-va: implemented

src/bannerService.ts
dmekala-va: implemented

src/bannerService.ts
johnmweisz: Ok we can leave as is, thanks for checking.

src/bannerService.ts
johnmweisz: Can different entries have different TTL in the same table?  If so we definitely need this to last more than 3600s.   Otherwise we might need a second table.

src/bannerService.ts
johnmweisz: We can leave it how it is for now I’m more worried about ttl

src/bannerService.ts
dmekala-va: Updated the code(createBanner and updateBanner) to use TTL for 7 days. I used the set to alter this value, I think this way works

src/bannerService.ts
```diff
@@ -0,0 +1,91 @@
+import Banner from "./banner";
+import { ICache } from "./routes/types";
+
+export interface BannerService {
+  getBanner(id: string): Promise<Banner | null>;
+  getBanners(): Promise<Banner[]>;
+  createBanner(banner: Banner): Promise<Banner | null>;
+  updateBanner(id: string, banner: Banner): Promise<boolean>;
+  deleteBanner(id: string): Promise<boolean>;
```
mwise-va, line 9: Should we add a delete banners as well? Similar to the get all banners. This way we won't have to perform a GET -> Del -> Del if there are more then 1 banner at once. We could just run a Del all banners

src/banner.ts
```diff
@@ -1,16 +1,16 @@
...
   message: string;
   enabled: boolean;
   priority: number;
   order: number;
 
   constructor(
     id: string,
-    startTime: Date,
-    endTime: Date,
+    startTime: string,
```
johnmweisz, line 12: Sorry I should have been more specific, this can be a unix timestamp aka a number, that will make frontend display logic simpler, though that is subject to change.  I think we start with unix and adjust as we build out the frontend.

src/banner.ts
```diff
@@ -1,16 +1,16 @@
...
+    endTime: string,
```
johnmweisz, line 13: number

src/bannerService.ts
johnmweisz: nice, thanks

src/bannerService.ts
johnmweisz: I don't think we will be using delete all much since it will have a display start and end and automatically delete after TTL, I'm ok with leaving this out for now.

dmekala-va: We don't have to create a new function to do that. We can use the existing keys endpoint to delete all banners similar to how we get all banners by the "banner:*" pattern. Just not sure we want to add something like that though

src/banner.ts
```diff
@@ -0,0 +1,35 @@
+export default class Banner {
+  id: string;
+  startTime: number;
+  endTime: number;
+  message: string;
+  enabled: boolean;
+  priority: number;
```
mwise-va, line 7: Just curious, why do we have a priority and order?  I know it was stated in the ticket, but I could see this done with just order and not priority. There would be a check to see if that position (order) is already taken or not.

johnmweisz: I think order designates physical placement, while priority will designate a UI style.  The name is a bit confusing, perhaps we can think of something else.

johnmweisz: Maybe something like 'alertStyle' see https://getbootstrap.com/docs/4.0/components/alerts/ as an example.

mwise-va: What do you mean by priority will designate a UI style? Wouldn't that be done with CSS etc. 

A similar feature Okta does for adding policies, is Priority is used, but order is not. Just don't see added value from having both.

src/banner.ts
```diff
@@ -0,0 +1,35 @@
...
+  alertStyle: number;
+  order: number;
+
+  constructor(
+    id: string,
+    startTime: number,
+    endTime: number,
+    message: string,
+    enabled: true,
+    priority: 0,
```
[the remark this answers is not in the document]
dmekala-va: updated

johnmweisz (approved):

mwise-va (approved):

PR #493, opened 2024-07-09
API-37717 IDP Metadata loads from Config

mwise-va:
## Description
To make the retrieving and loading of IDP metadata more stable prior application start up, saml-proxy will no longer make http request to retrieve IDP metadata before application start up.  Instead, saml-proxy will pull the IDP metadata from configs. 
## Related Issue
https://jira.devops.va.gov/browse/API-37717


johnmweisz:
fyi this is blocked until config is updated.

mwise-va:
Closing, since we are going with the fallback from configs approach.

PR #494, opened 2024-07-12
API-37717 IDP Metadata loads from Config as Fallback

mwise-va:
## Description
To make the retrieving and loading of IDP metadata more stable prior application start up, saml-proxy will first try to make an http request to retrieve IDP metadata before application start up.  If that fails, then saml-proxy will pull the IDP metadata from configs. 
## Related Issue
https://jira.devops.va.gov/browse/API-37717

### DIFF
--- a/src/app.js
+++ b/src/app.js
@@ -72,7 +72,7 @@ const handleMetadata = (argv) => {
  */
 function runServer(argv) {
   const strategies = new Map();
-  IdPMetadata.fetch(argv.spIdpMetaUrl)
+  IdPMetadata.fetch(argv.spIdpMetaUrl, argv.spIdpMetadata)
     .then(handleMetadata(argv))
     .then(() => {
       const app = express();
@@ -83,7 +83,7 @@ function runServer(argv) {
       app.use(passport.initialize());
       if (argv.idpSamlLoginsEnabled) {
         argv.idpSamlLogins.forEach((spIdpEntry) => {
-          IdPMetadata.fetch(spIdpEntry.spIdpMetaUrl)
+          IdPMetadata.fetch(spIdpEntry.spIdpMetaUrl, spIdpEntry.spIdpMetadata)
             .then(handleMetadata(spIdpEntry))
             .then(() => {
               spIdpEntry.spKey = argv.spKey;
--- a/src/idpMetadata.js
+++ b/src/idpMetadata.js
@@ -52,7 +52,7 @@ function getFirstCert(keyEl) {
 }
 
 /**
- * Function to read and parse XML file
+ * Read and parse XML file
  *
  * @param {string} filePath Path to the XML file
  * @returns {*} Parsed XML object
@@ -118,7 +118,7 @@ export function fetch(url, idpMetadataFallback) {
 }
 
 /**
- * Process IDP metadata from XML data
+ * Process IDP XML metadata
  *
  * @param {*} docEl Parsed XML object
  * @param {*} metadata Metadata object to populate
--- a/src/idpMetadata.js
+++ b/src/idpMetadata.js
@@ -1,8 +1,15 @@
 "use strict";
-
+const fs = require("fs");
 const xml2js = require("xml2js");
-const logger = require("./logger");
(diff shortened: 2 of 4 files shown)


src/idpMetadata.js
```diff
@@ -43,118 +50,150 @@ function getFirstCert(keyEl) {
...
+            "urn:oasis:names:tc:saml:2.0:bindings:http-redirect"
+          );
+          metadata.slo.postUrl = getBindingLocation(
+            ssoEl.SingleLogoutService,
+            "urn:oasis:names:tc:saml:2.0:bindings:http-post"
+          );
         }
-      })
-      .catch(() => {
-        logger.error("Error receiving metadata");
```
johnmweisz, line 157: do we still log this error somewhere? we would want to know if the get request fails and if the xml parsing fails, because they indicate different things.

src/app.js
```diff
@@ -72,7 +72,7 @@ const handleMetadata = (argv) => {
  */
 function runServer(argv) {
   const strategies = new Map();
-  IdPMetadata.fetch(argv.spIdpMetaUrl)
+  IdPMetadata.fetch(argv.spIdpMetaUrl, argv.spIdpMetadata)
```
johnmweisz, line 75: fyi might want to update the dev config base with new arg

src/idpMetadata.js
mwise-va: Added logging for if the get request fails and if the xml parsing fails

src/idpMetadata.js
```diff
@@ -43,118 +50,154 @@ function getFirstCert(keyEl) {
...
-                );
-                metadata.slo.postUrl = getBindingLocation(
-                  ssoEl.SingleLogoutService,
-                  "urn:oasis:names:tc:saml:2.0:bindings:http-post"
-                );
-              }
-            }
+        parser.parseString(responseData, (err, docEl) => {
+          if (err || !docEl) {
+            console.error(
```
johnmweisz, line 100: logger.error

src/idpMetadata.js
```diff
@@ -43,118 +50,154 @@ function getFirstCert(keyEl) {
...
+              .then((xmlData) => processMetadata(xmlData, metadata))
+              .then(resolve)
+              .catch(reject);
+          } else {
+            processMetadata(docEl, metadata).then(resolve).catch(reject);
+          }
+        });
+      })
+      .catch(() => {
+        console.error(
```
johnmweisz, line 113: logger.error

src/idpMetadata.js
```diff
@@ -1,8 +1,15 @@
 "use strict";
-
+const fs = require("fs");
 const xml2js = require("xml2js");
-const logger = require("./logger");
```
johnmweisz, line 4: this should stay

src/idpMetadata.js
mwise-va: is there a reason, to use logger.error vs console.error? 

I see the same results with console.error in outputting when an error occurs. Using logger.error would require me to make more code changes

src/idpMetadata.js
mwise-va: logger.error would require me to make more code changes, prefer using console.error if I can

src/idpMetadata.js
mwise-va: Something to consider, is that this feature is made to have a fallback. console.error has been killing the application and not allowing for a fallback. This is the reason I chose not to use it

src/idpMetadata.js
mwise-va: logger.error was not allowing a fallback approach and would stop the application

mwise-va:
Bug ticket to address `logger.error` abruptly stopping saml-proxy prior to application start up - https://jira.devops.va.gov/browse/API-38222

johnmweisz (commented):
I think we need to add the new config item (spIdpMetadata) to the cli index page to keep that in sync

johnmweisz (approved):

crolarlibertyva (approved):

PR #495, opened 2024-07-30
API-36891 Update login page

dmekala-va:
This PR is to update the saml-proxy login page to reflect the UX updates to the design of the page. 

Link to Jira: https://jira.devops.va.gov/browse/API-36891


package.json
```diff
@@ -70,7 +70,8 @@
     "node": ">=6.0.0"
   },
   "dependencies": {
-    "@department-of-veterans-affairs/formation": "^8.0.1-3",
+    "@department-of-veterans-affairs/component-library": "^43.0.1",
```
johnmweisz, line 73: Since we are not using these changes (css file shows no changes and we are not importing the component css) these package file changes can be reverted.

johnmweisz: I plan to have a follow up story upgrading this package and using the component library when we have more time.  (This will translate over to any new version of this page so I think worth it but lower priority.)

mwise-va:
Tested on all dimension sizes for phones within the google chrome dev tools. Looks good

I think the centering is off for the content under the VA logo (looks a little to the left) but UX can decide on that stuff

mwise-va (approved):

crolarlibertyva (approved):

PR #504, opened 2024-08-15
switch idp for regression test to login-gov

TylerGuerrero-va:
Draft PR for changing IDP for the regression tests for SAML-Proxy to Login-Gov for better stability

1) Two of the test's 'ICN Errors' and 'modify' are currently failing right now
- With The 'ICN Errors' test its failing right now in the sandbox env but not the dev env. In sandbox  its still able to get an access token and redirect the url even when using the user: va.api.user+007@gmail.com with the faulty ICN number, I did a manually test with Postman with that user in the dev env and did get the correct output, the expected Match Error page but in sandbox I am able to get the access token and it executes the redirect, so not sure why there is different behavior depending on the environment, Going to do further poking around to see what's going on with mpi

-  The 'modify' test I am still trying to figure out why its failing

2) Link to sh file with export values to run the regression test locally: keybase://team/lighthouse_pivot/configs/saml-proxy/regression-test-envs.sh

3) Added the new env vars to the param store to align with the regression tests, kept the old idme ones for history but added the new login-gov ones:  [SAML-proxy param store](https://us-gov-west-1.console.amazonaws-us-gov.com/systems-manager/parameters?region=us-gov-west-1&tab=Table#list_parameter_filters=Name:Contains:%2Fdvp%2Fcommon%2Fecs-saml-proxy%2Fcodebuild)


crolarlibertyva (approved):
Works as well as the previous ID me tests used to work. Maybe will try to recover the old 043 user for login.gov. I think that account still exists.

crolarlibertyva:
Before you can move this from draft, you need to align this with the new environment variables.
 - https://github.com/department-of-veterans-affairs/lighthouse-saml-proxy/blob/API-38828-regression-tests-login-gov/cicd/buildspec-deploy.yml

crolarlibertyva (changes requested):
The Makefile needs updating too.
 - https://github.com/department-of-veterans-affairs/lighthouse-saml-proxy/blob/d69c60a6a34d31ea330a13458ee225cad6bea12d/Makefile#L106-L118

crolarlibertyva (changes requested):
The entrypoint_test.sh needs adjusting as well.

TylerGuerrero-va:
Would have to do it to the entry point_test.sh right ?

crolarlibertyva (approved):

mwise-va (approved):

mwise-va:
Tested locally, all passing but modify test

  Regression tests
    ✓ Happy Path (12799 ms)
    ✓ ICN Error (11394 ms)
    ✓ Replay (12419 ms)
    ✓ Empty SSO (2738 ms)
    ✓ Empty Assertion Response (2708 ms)
    ✓ 404 (2731 ms)
    ✕ modify (71101 ms)

PR #509, opened 2024-10-10
API-41014 Fraud Identity Theft

dmekala-va:
This PR is to add the fraud identity theft check in saml-proxy to block any login via the SAML Proxy that is identified to have a Fraud Identity Flag returned from MPI/ICN.

Link to Jira: https://jira.devops.va.gov/browse/API-41014

### DIFF
--- a/src/MpiUserClient.test.ts
+++ b/src/MpiUserClient.test.ts
@@ -50,6 +50,7 @@ beforeEach(() => {
             icn: "fakeICN",
             first_name: "Edward",
             last_name: "Paget",
+            id_theft_indicator: true,
           },
         },
       },
@@ -62,7 +63,8 @@ describe("getMVITraitsForLoa3User", () => {
     const client = new MpiUserClient(
       "faketoken",
       "https://example.gov/mvi-user",
-      "faketoken"
+      "faketoken",
+      true
     );
     await client.getMpiTraitsForLoa3User(samlTraitsEDIPI);
     expect(axios.post).toHaveBeenCalledWith(
@@ -89,7 +91,8 @@ describe("getMVITraitsForLoa3User", () => {
     const client = new MpiUserClient(
       "faketoken",
       "https://example.gov/mpi-user",
-      "faketoken"
+      "faketoken",
+      true
     );
     await client.getMpiTraitsForLoa3User(samlTraitsICN);
     expect(axios.post).toHaveBeenCalledWith(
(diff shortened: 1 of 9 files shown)


src/MpiUserClientConfig.js
```diff
@@ -3,5 +3,6 @@ export default class MpiUserClientConfig {
     this.mpiUserEndpoint = argv.mpiUserEndpoint;
     this.accessKey = argv.accessKey;
     this.apiKey = argv.vetsAPIToken;
+    this.fraudIdTheft = argv.fraudBlockEnabled || false;
```
johnmweisz, line 6: probably want to keep the config boolean implication in this name so instead of this.fraudIdTheft maybe this.fraudBlockEnabled

src/cli/index.js
```diff
@@ -41,6 +41,13 @@ export function processArgs() {
           KEY_CERT_HELP_TEXT
         ),
       },
+      fraudBlockEnabled: {
+        description:
+          "Enable or disable blocking logins based on the fraud identity indicator",
+        required: false,
+        boolean: true,
+        default: true,
```
johnmweisz, line 49: this can default false, we want the behavior that if the config item is not set the block is disabled

src/MpiUserClient.test.ts
```diff
@@ -50,6 +50,7 @@ beforeEach(() => {
             icn: "fakeICN",
             first_name: "Edward",
             last_name: "Paget",
+            id_theft_indicator: true,
```
johnmweisz, line 53: can we also get a test where the mpi response idtheftindicator is true but the block setting is false and confirm it doesn't block?

johnmweisz:
can also update the default config example with this setting added

src/MpiUserClient.test.ts
dmekala-va: Will add this to the acsHandler since there are similar tests like that already there

crolarlibertyva (approved):
```
{
  "level": "warn",
  "message": "Fradulent identity detected, blocking login.",
  "request_id": "565fec10-87fc-11ef-9e04-8fef16314a83",
  "service": "saml-proxy",
  "time": "2024-10-11T18:12:13.428Z"
}
```

mwise-va (dismissed):

TylerGuerrero-va (dismissed):

src/rsa.json
crolarlibertyva, line 1: Is this an accident?

dmekala-va: removed

crolarlibertyva (approved):

mwise-va (approved):

PR #517, opened 2024-11-12
API-41180 Horizontal width same size, by removing padding

mwise-va:
## Description
Horizontal width same size, by removing padding.


johnmweisz (approved):

TylerGuerrero-va:
Gonna add this to Identity-Proxy

